PR #12, opened 2020-10-07
fix: test compilation issues on Xcode 12

imaginaris:
# Description
I had some "Illegal statement" errors on Xcode 12 while trying to compile Test target.
According to this thread: https://github.com/Quick/Nimble/issues/809
it seems like it's some kind of Swift regression so as a workaround all `expect` calls with trailing closures have been wrapped in parentheses.

## Links
https://github.com/Quick/Nimble/issues/809

# Checklist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: swift
 xcode_workspace: Example/InAppMessaging.xcworkspace
 xcode_scheme: InAppMessaging-Example
-osx_image: xcode11.5
+osx_image: xcode11.7
 
 script:
   - bundle exec fastlane ci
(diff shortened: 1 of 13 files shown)


donnie-jp (approved):
LGTM but do we need to run the CI on Xcode 12?
Can we wait until the Quick/Nimble issues are fixed?
I can go with whatever you prefer.

imaginaris:
> LGTM but do we need to run the CI on Xcode 12?
> Can we wait until the Quick/Nimble issues are fixed?
> I can go with whatever you prefer.

Yeah. I wanted to check if everything is ok but it seems that there are more issues with Xcode 12 :/ (i.e. xcov) so I changed the CI setting to Xcode 11.7

Issue #10, opened 2020-10-05
SDKCF-2871


donnie-jp:
Requirements
WHAT: Investigate the implementation of a similar feature than in Android to register/unregister Activity (or Fragment)

WHY:
To prevent timing issue where a campaign could be displayed on a non-related screen (Activity/Fragment).

Campaign set to be displayed on screen A
User reach screen A and directly reach screen B
Campaign didn't had time to be displayed on screen A but gets displayed on screen B

Problem: This campaign was a tutorial for screen A

Output from the Investigation
* Confirmation if this improvement can be done
* Decide on implementation
* Sync with Android to align SDKs logic
* Draft spec (if needed)
* Create related implementation

donnie-jp:
Explanation shared by Vincent:

> Here is the summary of my understanding regarding the IAM issue:
> 
> They are triggering the event in Home Tab - they expect to display the campaign here
> But in some instances, there is a quick transition to another tab
> So 1 out of 5 (I guess), the campaign is displayed in the other tab
> Based on our specs/behavior, this is an expected behavior
> It is just the timing of the campaign/event processing
> 
> In Android, this can be handled by register and unregister API
> 
> An activity can be registered if it will allow IAM to be displayed, then it should be unregister when transitioning to another activity
> 
> They are using fragments for the tab which will require investigation to handle.
> In iOS, I’m not sure if there is a way to this with the same logic in Android

imaginaris:
Proposed solution:

1. Extend SDK's API method `logEvent` by adding a parameter `context` of new type `EventContext`
2. Add new optional delegate with method `inAppMessagingShouldShowCampaignMessage` that will accept campaign's title and contexts from logged events as a parameter and expect a bool value in return

Example:
A host app logs an event with context named "tutorial". When a campaign triggered by that event is about to be displayed it will call the new delegate method with "tutorial" context to confirm or abort displaying the message.

donnie-jp:
Closed in https://github.com/rakutentech/ios-inappmessaging/pull/13

PR #13, opened 2020-10-07
feat: add Context to logged events and validate them before displaying campaign message (SDKCF-2871)

imaginaris:
# Description
Solution no. 2 for issue https://github.com/rakutentech/ios-inappmessaging/issues/10
Tested on a iOS 14 simulator using Example project ~~and mocked response from the server.~~ with a campaign crated in PROD dashboard.

## Links
SDKCF-2871
https://github.com/rakutentech/ios-inappmessaging/issues/10

# Checklist


RInAppMessaging/Classes/Protocols/TaskSchedulable.swift
```diff
@@ -14,10 +14,11 @@ internal protocol TaskSchedulable: AnyObject {
 /// Default implementation.
 extension TaskSchedulable {
     func scheduleWorkItem(milliseconds: Int, wallDeadline: Bool, closure: @escaping () -> Void) {
-
-        scheduledTask?.cancel()
-        scheduledTask = WorkScheduler.scheduleTask(milliseconds: milliseconds,
-                                                   closure: closure,
-                                                   wallDeadline: wallDeadline)
+        DispatchQueue.main.async {
```
donnie-jp, line 17: Why put this on the main thread?

RInAppMessaging/Classes/RInAppMessaging.swift
```diff
@@ -1,3 +1,9 @@
+/// Protocol for optional delagate
+@objc public protocol RInAppMessagingDelegate: AnyObject {
+    /// Method called only for campaigns with context just before displaying its message
+    func inAppMessagingShouldShowCampaignsWithContexts(contexts: [String], campaignTitle: String) -> Bool
```
donnie-jp: `Campaigns` or `Campaign` ?
Cos there's `campaignTitle` which suggests singular campaign.

RInAppMessaging/Classes/Router.swift
```diff
@@ -64,7 +70,7 @@ internal class Router: RouterType {
             }
 
             DispatchQueue.main.async {
-                guard let view = viewConstructor?() else {
+                guard let view = viewConstructor?(), confirmation() else {
```
donnie-jp: Would `confirmation() == true` be clearer?

RInAppMessaging/Classes/RInAppMessaging.swift
donnie-jp: Why are there multiple contexts? To provide flexibility?

RInAppMessaging/Classes/Models/Responses/PingResponse.swift
```diff
@@ -27,6 +27,11 @@ internal struct Campaign: Decodable, Hashable {
         let endTimeMilliseconds = data.messagePayload.messageSettings.displaySettings.endTimeMilliseconds
         return endTimeMilliseconds < Date().millisecondsSince1970
     }
+    var contexts: [String] {
+        data.messagePayload.title.components(separatedBy: "]").dropLast().map { substring in
```
imaginaris, line 31: I tried to use regex but with creation of NSRegularExpression object the current solution is faster

RInAppMessaging/Classes/Protocols/TaskSchedulable.swift
imaginaris: It just to prevent resource conflicts when accessing scheduledTask property. The issue was visible sometimes when running tests (EXC_BAD_ACCESS crash)

RInAppMessaging/Classes/Models/Responses/PingResponse.swift
donnie-jp: yeah this hurt my head a bit 😄

RInAppMessaging/Classes/RInAppMessaging.swift
imaginaris: yes. Single campaign, multiple contexts

RInAppMessaging/Classes/Router.swift
imaginaris: I wanted to put a name for this closure so it wouldn't be necessary but I couldn't find a better name. Probably `confirmation() == true` is clearer

RInAppMessaging/Classes/InAppMessagingModule.swift
```diff
@@ -99,6 +100,14 @@ extension InAppMessagingModule {
             self.campaignsListManager.refreshList()
         }.perform()
     }
+
+    func shouldShowCampaignMessage(title: String, contexts: [String]) -> Bool {
+        guard let delegate = delegate else {
+            return true
+        }
+        return delegate.inAppMessagingShouldShowCampaignsWithContexts(contexts: contexts,
```
donnie-jp: Campaigns -> Campaign

donnie-jp: Do you need a respondsToSelector check here? You've checked the delegate exists but we don't know if the delegate method was implemented.

Tests/ReadyCampaignDispatcherTests.swift
```diff
@@ -113,6 +170,20 @@ class ReadyCampaignDispatcherTests: QuickSpec {
                         dispatcher.dispatchAllIfNeeded()
                         expect(delegate.wasPingCalled).toEventually(beTrue())
                     }
+
+                    it("won't call shouldShowCampaignMessage delegate method for non-test campaigns") {
+                        let campaign = TestHelpers.generateCampaign(id: "test", title: "[ctx] title")
+                        dispatcher.addToQueue(campaign: campaign)
+                        dispatcher.dispatchAllIfNeeded()
+                        expect(delegate.wasShouldShowCalled).toAfterTimeout(beFalse())
```
donnie-jp, line 178: Why isn't `shouldShowCampaignMessage` called if it's a non-test campaign?

Tests/SerializationTests.swift
```diff
@@ -17,5 +17,45 @@ class SerializationTests: QuickSpec {
...
+
+        describe("Campaign model") {
+            context("When parsing title for contexts") {
+
+                it("will return empty array if there are no contexts") {
+                    let campaign = TestHelpers.generateCampaign(id: "id", title: "title")
+                    expect(campaign.contexts).to(beEmpty())
+                }
+
+                it("will properly read context if there's nothing beside it") {
```
donnie-jp: "will properly read context when it is the only string content"

donnie-jp (changes requested):
Looks good. I really like the solid tests.
I left a few comments/questions.

RInAppMessaging/Classes/InAppMessagingModule.swift
imaginaris: This method is not available for Swift objects. I think all swift protocols are marked as `required` in ObjC

Tests/ReadyCampaignDispatcherTests.swift
imaginaris: In this context `when display permission is denied` it shouldn't. It would be pointless to ask host app about campaign that won't be displayed anyway

Tests/SerializationTests.swift
imaginaris: Much better 👍

Tests/ReadyCampaignDispatcherTests.swift
donnie-jp: Ah ok, I missed that context.

donnie-jp (approved):

donnie-jp:
Please test using a real IAM portal campaign before merging.

RInAppMessaging/Classes/InAppMessagingModule.swift
donnie-jp: Ok, I thought because of the `@objc` it would be available but I now see the delegate extends `AnyObject`.

RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -75,14 +76,24 @@ internal class CampaignDispatcher: CampaignDispatcherType {
                     Error: Campaign (\(campaign.id)) does not exist in the repository anymore (race condition?). Proceeding with old data...
                     """)
             }
-            self.router.displayCampaign(campaign) { [weak self] in
+
+            let campaignTitle = campaign.data.messagePayload.title
+            self.router.displayCampaign(campaign, confirmation: {
+                let contexts = campaign.contexts
+                guard let delegate = self.delegate, !contexts.isEmpty, !campaign.data.isTest else {
```
malayli, line 83: What is `isTest`property?

RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -75,14 +76,24 @@ internal class CampaignDispatcher: CampaignDispatcherType {
...
+                    return true
+                }
+
+                return delegate.shouldShowCampaignMessage(title: campaignTitle,
+                                                          contexts: contexts)
+            }, completion: { [weak self] in
                 guard let strongSelf = self else {
                     return
                 }
```
malayli, line 92: You can use this instead:
guard let \`self\` = self else { return }

imaginaris:
> Please test using a real IAM portal campaign before merging.

Done

RInAppMessaging/Classes/CampaignDispatcher.swift
imaginaris: But then I'd have to use ``self` as a reference ?

RInAppMessaging/Classes/Router.swift
```diff
@@ -6,8 +6,11 @@ internal protocol RouterType: AnyObject {
     /// Contains logic to display the correct view type and create
     /// a view controller to present a single campaign.
     /// - Parameter campaign: The campaign object to display.
+    /// - Parameter confirmation: A handler called just before displaying.
     /// - Parameter completion: Completion handler called once displaying has finished.
-    func displayCampaign(_ campaign: Campaign, completion: @escaping () -> Void)
+    func displayCampaign(_ campaign: Campaign,
+                         confirmation: @escaping () -> Bool,
```
malayli: I think that you can add `@autoclosure`here so instead of having:
`confirmation: { true }` or `confirmation: { false }`
We would have:
`confirmation: true` or `confirmation: false`

But that would mean to call a function instead of a closure for a complex code.

imaginaris: I had a problem with complex code in CampaignDispatcher but I solved it just by adding `()` at the and so yeah I think I like this better now!

malayli:
@imaginaris This PR looks good. I just have few comments. When the PR is updated, I'll validate this PR.

RInAppMessaging/Classes/CampaignDispatcher.swift
imaginaris: to mark test campaigns. It's true when you use "Send to test device" button in the dashboard

malayli (approved):

RInAppMessaging/Classes/CampaignDispatcher.swift
imaginaris: Ok I'll use `guard let self = self else { return }` since it's possible since Swift 4.2 👍

PR #19, opened 2020-12-09
improve: update README to clarify user targeting and add corresponding assert in preference builder (SDKCF-3062)

imaginaris:
## Links
SDKCF-3062

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,13 +11,13 @@ In-App Messaging (IAM) module allows app developers to easily configure and disp
 RInAppMessaging SDK is distributed as a Cocoapod.  
 More information on installing pods: [https://guides.cocoapods.org/using/getting-started.html](https://guides.cocoapods.org/using/getting-started.html)
 
-1) Include the following in your application's Podfile
+1. Include the following in your application's Podfile
 
 ```ruby
 pod 'RInAppMessaging'
 ```
 
-2) Run the following in the terminal
+2. Run the following in the terminal
 
 ```
 pod install
@@ -27,7 +27,7 @@ pod install
 
 **Note:** Currently we do not host any public APIs but you can create your own APIs and configure the SDK to use those.
 
-1)  To use the module you must set the following values in your app's `Info.plist`:
+To use the module you must set the following values in your app's `Info.plist`:
 
 | Key     | Value     |
 | :---:   | :---:     |
@@ -59,7 +59,7 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
 
 **Objective-C**
 
-```swift
+```objectivec
 @import RInAppMessaging;
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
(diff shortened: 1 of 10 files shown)


RInAppMessaging/Classes/IAMPreferenceBuilder.swift
```diff
@@ -27,6 +27,10 @@
 
     @objc @discardableResult
     public func build() -> IAMPreference {
+        if BundleInfo.applicationId?.starts(with: "jp.co.rakuten") == true, Bundle.tests == nil {
+            assert(preference.accessToken == nil || preference.accessToken != nil && preference.userId != nil,
```
malayli, line 31: About `preference.accessToken != nil && preference.userId != nil`, it means that it crashes if accessToken is not nil and userId is not nil, right?
Maybe the condition could be `preference.accessToken != nil && preference.userId == nil`?
So It crashes if the userId is nil even if there is a non nil accessToken?

imaginaris: The condition should say that "if you specify accessToken you must also specify userId". The following combinations of (userId, accessToken, result) should apply (nil, value, crash), (nil, nil, ok), (value, nil, ok), (value, value, ok). Assert crashes if condition returns false so you condition seems to do the opposite or maybe I misunderstand something.

malayli: Ok I understand the condition now. Thanks.

malayli (approved):

README.md
```diff
@@ -59,7 +59,7 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
 
 **Objective-C**
 
-```swift
+```objectivec
```
donnie-jp: Maybe we can remove the ObjC code examples now.

README.md
```diff
@@ -241,3 +243,8 @@ func application(_ application: UIApplication, didFinishLaunchingWithOptions lau
 * Run `bundle install` then run `bundle exec pod install`
 * Open `RInAppMessaging.xcworkspace` in Xcode then build/run
 * To run the tests press key shortcut command-U
+
+## **Troubleshooting**
+
+* (For our internal users) Getting error 401 when calling ping/impression endpoint - check your app's scopes list if it contains all required entries OR stop putting accessToken in `IAMPreference` (breaks user targeting)
```
donnie-jp: Instead of this, lets say
> Rakuten developers experiencing problems should refer to the Troubleshooting Guide on the internal developer documentation portal.

so that we keep troubleshooting info in one place and can combine it with backend info.

README.md
```diff
@@ -144,11 +144,13 @@ A preference is what will allow IAM to identify users for targeting and segmenta
 3.  AccessToken - This is the token provided by the internal RAuthentication SDK as the "accessToken" value
 
 To help IAM identify users, please set a new preference every time a user changes their login state i.e. when they log in or log out.  
-Not all identifiers have to be provided.
+During log out please call  `registerPreference()` with nil parameter instead of  `IAMPreference` object with nil properties.
```
donnie-jp: I don't think we need "instead of  `IAMPreference` object with nil properties".

Should it be "After logout is complete" rather than "During logout" ?

README.md
```diff
@@ -80,7 +80,7 @@ func logEvent(_ event: Event)
 
 IAM provides three pre-defined event types and a custom event type:
 
-1.  `AppStartEvent` - This event should be logged when the application is considered started by the host app. E.G AppDelegate's didFinishLaunchingWithOptions.
+1.  `AppStartEvent` - This event should be logged when the application is considered started by the host app. E.G AppDelegate's didFinishLaunchingWithOptions. It is persistent, meaning, once it's logged it will always satisfy corresponding trigger in a campaign. All subsequent logs of this event are ignored. Campaigns that require only AppStartEvent are shown once per app session.
```
donnie-jp: Should we define app session or is it too obvious?

RInAppMessaging/Classes/IAMPreferenceBuilder.swift
donnie-jp, line 31: Nice idea to add this.
Can you confirm that this assert is only executed in a debug build?

donnie-jp (changes requested):
Comments inline

RInAppMessaging/Classes/IAMPreferenceBuilder.swift
imaginaris: Can I confirm by definition?
https://developer.apple.com/documentation/swift/1541112-assert
`...active during testing but do not impact performance of shipping code.`

README.md
imaginaris: I thought about that. Maybe it's too vague. I'll try to be more specific

README.md
imaginaris: Right "during logout" can be difficult :D I wanter do be sure that no one thinks of an idea to pass an empty object instead of nil but I think you are right that it's covered by previous part of the sentence

donnie-jp (approved):
LGTM

PR #24, opened 2020-12-23
ci: add bitrise.yml

imaginaris:
# Description
CI config will be moved from web to `bitrise.yml` file

# Checklist

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -1,0 +1,138 @@
+---
+format_version: '8'
+default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
+project_type: ios
+trigger_map:
+- push_branch: "*"
+  workflow: primary
+- pull_request_source_branch: "*"
+  workflow: primary
+- tag: "*.*.*"
+  workflow: deploy
+workflows:
+
+  primary:
+    before_run:
+    - _setup
+    after_run:
+    - _report
+    steps:
+    - fastlane@2:
+        inputs:
+        - update_fastlane: 'false'
+        - lane: ci
+        title: Build & Test
+
+  deploy:
+    before_run:
+    - _setup
+    steps:
+    - script@1:
+        title: Verify branch
+        inputs:
+        - content: |-
+            #!/usr/bin/env bash
+            set -ex
+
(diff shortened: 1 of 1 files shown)


donnie-jp (approved):
LGTM

donnie-jp:
> ## Danger has errored
> [!] Invalid `Dangerfile` file: �[31mError creating your coverage report - see the log above�[0m

That's weird cos in the bitrise log it looks like coverage was generated OK.

imaginaris:
> > ## Danger has errored
> > [!] Invalid `Dangerfile` file: �[31mError creating your coverage report - see the log above�[0m
> 
> That's weird cos in the bitrise log it looks like coverage was generated OK.

It's because that damned javascript unit test is still failing on Travis even after I updated the xcode version.
I think we should disable Travis for now. because now it creates 2 codecov reports, 2 danger reports and it'll also deploy 2 times if a tag is pushed.
We could just disable the webhook. What do you think @donnie-jp

donnie-jp:
> We could just disable the webhook. What do you think @donnie-jp

Yes I agree.
I've disabled "Build pushed branches" and "Build pushed pull requests" in travis so I think it won't run in future.

bitrise.yml
```diff
@@ -0,0 +1,138 @@
...
+    BITRISE_SCHEME: RInAppMessaging-Example
+  - opts:
+      is_expand: false
+    BITRISE_EXPORT_METHOD: ad-hoc
+  - opts:
+      is_expand: false
+    CODECOV_TOKEN: 4144ada4-65b7-44bd-b478-48e3583a55cd
+  - opts:
+      is_expand: false
+    DANGER_GITHUB_API_TOKEN: 78fb7ddf8a020d492ed125f3277d631b3a5b7e6c
```
donnie-jp: GitHub revoked this token due to being included in this commit.
It will need to be injected as an env var from the bitrise web config. I generated a new token and added it to the accounts page.

imaginaris: Oh noo... :( This mean we can't use repo based bitrise.yml. The best we can do is to use that workaround we have in our other projects - that way we'll be able to keep env vars on the web and workflows in the repo.

PR #49, opened 2021-02-05
improve: change the default status bar overlay color in dark mode (SDKCF-3303)

imaginaris:
## Links
SDKCF-3303

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/Extensions/UIApplication+IAM.swift
+++ b/RInAppMessaging/Classes/Extensions/UIApplication+IAM.swift
@@ -21,7 +21,15 @@ extension UIApplication {
                 .filter({ $0.activationState == .foregroundActive })
                 .compactMap({ $0 as? UIWindowScene })
                 .first {
-            return keyScene.statusBarManager?.statusBarStyle
+
+            var style = keyScene.statusBarManager?.statusBarStyle
+            // `default` style doesn't tell us anything, so we can fallback to the most common pattern:
+            // light status bar content in dark mode and dark content in light mode.
+            if style == .default, let keyWindow = getKeyWindow() {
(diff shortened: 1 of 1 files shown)


RInAppMessaging/Classes/Extensions/UIApplication+IAM.swift
```diff
@@ -21,7 +21,15 @@ extension UIApplication {
                 .filter({ $0.activationState == .foregroundActive })
                 .compactMap({ $0 as? UIWindowScene })
                 .first {
-            return keyScene.statusBarManager?.statusBarStyle
+
+            var style = keyScene.statusBarManager?.statusBarStyle
+            // `default` style doesn't tell us anything, so we can fallback to the most common pattern:
+            // light status bar content in dark mode and dark content in light mode.
+            if style == .default, let keyWindow = getKeyWindow() {
+                let isDarkMode = keyWindow.traitCollection.userInterfaceStyle == .dark
```
donnie-jp, line 29: It seems like `userInterfaceStyle` is only available >= iOS 12?
https://developer.apple.com/documentation/uikit/uitraitcollection/1651063-userinterfacestyle

imaginaris: True, but `.darkContent` isn't (status bar style). I would have to use my own enum to support that piece of logic on iOS 12. Do you think it's worth it?

donnie-jp: From the diff context I couldn't see that this was already wrapped in an `#available(iOS 13.0, *)` check. I was worried about a runtime crash. I tried putting the `let isDarkMode` line into an app and it wouldn't even compile without an availability check - Swift FTW! 😅

donnie-jp (approved):
LGTM

PR #55, opened 2021-04-17
SDKCF-3654: [IAM][iOS] - Handle ‘429 too many requests’ response to Config/Ping APIs

malayli:
# Description
Retry logic when a 429 error is received for ping and getConfigData.
The initial value for retry is 60 seconds, then this value is exponentially increased + a random value is added.

## Tests
I tested getConfigData and Ping requests by mocking them on Charles Proxy and by returning 429 error:
HTTP/1.1 429

## Links
Add links to github/jira issues, design documents and other relevant resources (e.g. previous discussions, platform/tool documentation etc.)

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/Constants.swift
+++ b/RInAppMessaging/Classes/Constants.swift
@@ -40,4 +40,16 @@ internal enum Constants {
         static let impressions = "InAppMessaging_impressions"
         static let events = "InAppMessaging_events"
     }
+
+    enum Retry {
+        enum Default {
+            static let initialRetryDelayMS = Int32(10000)
+        }
+
+        enum TooManyRequestsError {
+            static let initialRetryDelayMS = Int32(60000)
+            static let backOffLowerBoundInSecond = Int32(1) // second
+            static let backOffUpperBoundInSecond = Int32(60) // second
+        }
+    }
 }
--- a/RInAppMessaging/Classes/Extensions/Int32+IAM.swift
+++ b/RInAppMessaging/Classes/Extensions/Int32+IAM.swift
@@ -1,0 +1,19 @@
+import Foundation
+
+extension Int32 {
+    /// Exponential backoff
+    mutating func increaseBackOff() {
+        self = multipliedReportingOverflow(by: 2).partialValue
+    }
+
+    /// Exponential backoff with a random value
+    ///
+    /// - Parameters:
+    ///     - min: the minimum value in second.
+    ///     - max: the maximum value in second.
+    mutating func increaseRandomizedBackoff(min: Int32 = Constants.Retry.TooManyRequestsError.backOffLowerBoundInSecond,
+                                            max: Int32 = Constants.Retry.TooManyRequestsError.backOffUpperBoundInSecond) {
+        increaseBackOff()
+        self += Int32.random(in: min...max)*1000
+    }
+}
--- a/Tests/ConfigurationManagerSpec.swift
+++ b/Tests/ConfigurationManagerSpec.swift
@@ -90,6 +90,13 @@ class ConfigurationManagerSpec: QuickSpec {
                         configurationService.simulateRequestFailure = false
                     }
                 }
+
+                it("should retry for .tooManyRequestsError error") {
+                    configurationService.simulateRequestFailure = true
+                    configurationService.mockedError = .tooManyRequestsError
+                    configurationManager.fetchAndSaveConfigData(completion: { _ in })
+                    expect(configurationManager.scheduledTask).toEventuallyNot(beNil())
+                }
             }
         }
     }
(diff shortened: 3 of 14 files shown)


RInAppMessaging/Classes/Extensions/Int32+IAM.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+
+extension Int32 {
+    /// Exponential backoff
+    mutating func increaseBackOff() {
+        self = multipliedReportingOverflow(by: 2).partialValue
+    }
+
+    /// Exponential backoff with a random value
+    mutating func increaseRandomedBackoff() {
```
donnie-jp: Randomed -> Randomized is better I think

malayli: Ok fixed.

donnie-jp (approved):
LGTM

donnie-jp (changes requested):
We won't merge to master - we'll merge to `pp-support` branch instead.

donnie-jp (approved):

RInAppMessaging/Classes/Extensions/Int32+IAM.swift
```diff
@@ -0,0 +1,14 @@
...
+extension Int32 {
+    /// Exponential backoff
+    mutating func increaseBackOff() {
+        self = multipliedReportingOverflow(by: 2).partialValue
+    }
+
+    /// Exponential backoff with a random value
+    mutating func increaseRandomizedBackoff() {
+        increaseBackOff()
+        self += Int32.random(in: 1..<60)
```
imaginaris: Those are milliseconds so it should be 1..<60000
Also could you put upper bound (60) in a separate var (static let even) to indicate what it is and also to clarify time unit.
Also I'm not sure if it isn't better to get upper bound as a parameter. That way we won't force this method to work only with MS

Tests/ConfigurationManagerSpec.swift
```diff
@@ -90,6 +90,13 @@ class ConfigurationManagerSpec: QuickSpec {
                         configurationService.simulateRequestFailure = false
                     }
                 }
+
+                it("should retry for .tooManyRequestsError error") {
+                    configurationService.simulateRequestFailure = true
+                    configurationService.mockedError = .tooManyRequestsError
+                    configurationManager.fetchAndSaveConfigData(completion: { _ in })
+                    expect(configurationManager.scheduledTask).toEventuallyNot(beNil())
```
imaginaris, line 98: Is there a way to kill this (and other) scheduled task after this evaluation? I didn;t think of it earlier

imaginaris (changes requested):

RInAppMessaging/Classes/Extensions/Int32+IAM.swift
donnie-jp: Can put it in the Constants.swift file I think

Tests/ConfigurationManagerSpec.swift
imaginaris: Ok. don't worry about it for now. I'll take care of it later

RInAppMessaging/Classes/Extensions/Int32+IAM.swift
malayli: Actually it is 1000..<60000 (between 1 second and 60 seconds)

malayli: Done.

RInAppMessaging/Classes/Constants.swift
```diff
@@ -40,4 +40,19 @@ internal enum Constants {
         static let impressions = "InAppMessaging_impressions"
         static let events = "InAppMessaging_events"
     }
+
+    enum Retry {
+        static let initialRetryDelayMS = Int32(10000)
+        static let initialRetryDelayForTooManyRequestsMS = Int32(60000)
```
imaginaris: Shouldn't that go to the enum below?

RInAppMessaging/Classes/Extensions/Int32+IAM.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+extension Int32 {
+    /// Exponential backoff
+    mutating func increaseBackOff() {
+        self = multipliedReportingOverflow(by: 2).partialValue
+    }
+
+    /// Exponential backoff with a random value
+    mutating func increaseRandomizedBackoff(min: Int32 = Constants.Retry.TooManyRequestsError.backOffLowerBoundMS,
```
imaginaris: 👍

imaginaris (approved):
LGTM. Left one small comment

RInAppMessaging/Classes/Extensions/Int32+IAM.swift
```diff
@@ -0,0 +1,15 @@
...
+                                            max: Int32 = Constants.Retry.TooManyRequestsError.backOffUpperBoundMS) {
+        increaseBackOff()
+        self += Int32.random(in: min..<max)
```
imaginaris: Hmm. I'm looking at the specs. Shouldn't that be [1...60] seconds i.e. `min...max`?

malayli: `retryDelayMS.increaseRandomizedBackoff()`
retryDelayMS is in millisecond.
But better if random is: `Int32.random(in: 1..<60)*1000`

imaginaris: Yeah, I'm not talking about the unit. I mean should we include upper bound in the range?
To change `min..<max` to `min...max`

malayli: Yes I agree. Fixed.

RInAppMessaging/Classes/Constants.swift
malayli: Done.

PR #56, opened 2021-04-18
feat: handle opt-out & max impression tracking logic solely on client side (SDKCF-3656)

imaginaris:
# Description
- Added cache to store ping response for each user
- Opt-out flag and impressionsLeft are now persisted for each campaign
- Fixed unit test issue when ping scheduler didn't have any delay and spawned multiple (mocked) requests

## Links
SDKCF-3656

# Checklist


RInAppMessaging/Classes/Repositories/CampaignRepository.swift
```diff
@@ -40,23 +45,37 @@ internal class CampaignRepository: CampaignRepositoryType {
...
         list.forEach { newCampaign in
             var updatedCampaign = newCampaign
             if let oldCampaign = oldList.first(where: { $0.id == newCampaign.id }) {
-                updatedCampaign = Campaign.updatedCampaign(updatedCampaign,
-                    asOptedOut: oldCampaign.isOptedOut)
+                updatedCampaign = Campaign.updatedCampaign(updatedCampaign, asOptedOut: oldCampaign.isOptedOut)
+
+                if retainImpressionsLeftValue {
+                    var newImpressionsLeft = oldCampaign.impressionsLeft
+                    let wasMaxImpressionsEdited = oldCampaign.data.maxImpressions != newCampaign.data.maxImpressions
```
imaginaris, line 68: I left this logic in case the backend will allow editing maxImpressions again

RInAppMessaging/Classes/Models/Responses/DisplayPermissionResponse.swift
```diff
@@ -1,4 +1,33 @@
...
     let display: Bool
     let performPing: Bool
+    let creationTimeMilliseconds: Int64 // Cache coding only
+
+    // For unit tests
+    init(display: Bool, performPing: Bool) {
+        self.display = display
+        self.performPing = performPing
+        creationTimeMilliseconds = Date().millisecondsSince1970
+    }
```
malayli, line 17: Can this init be declared in an extension in the unit tests target?

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
```diff
@@ -70,29 +89,50 @@ internal class CampaignRepository: CampaignRepositoryType {
...
+    func loadCachedData() {
+        guard let cachedData = userDataCache.getUserData(identifiers: preferenceRepository.getUserIdentifiers())?.campaignData else {
+            return
+        }
+        campaigns.set(value: cachedData)
+    }
+
+    private func findCampaign(withID id: String) -> Campaign? {
+        let list = campaigns.get()
+        return list.first(where: { $0.id == id })
```
malayli: Could be:
```
private func findCampaign(withID id: String) -> Campaign? {
    campaigns.get().first(where: { $0.id == id })
}
```

RInAppMessaging/Classes/UserDataCache.swift
```diff
@@ -0,0 +1,77 @@
...
+            let encodedData = try JSONEncoder().encode(cachedContainers)
+            userDefaults.set(encodedData, forKey: persistedDataKey)
+        } catch {
+            Logger.debug("UserDataCache encoding failed! \(error)")
+            assertionFailure()
+        }
+    }
+
+    private func userKey(from identifiers: [UserIdentifier]) -> Set<UserIdentifier> {
+        return Set<UserIdentifier>(identifiers)
```
malayli: You can remove return

Tests/Helpers/TestHelpers.swift
```diff
@@ -132,7 +132,7 @@ struct TestHelpers {
             }
 
             return PingResponse(
-                nextPingMilliseconds: 0,
+                nextPingMilliseconds: Int.max,
```
malayli, line 135: Does it mean that the next ping request will be sent after `Int.max` milliseconds?

RInAppMessaging/Classes/Models/Responses/DisplayPermissionResponse.swift
imaginaris: Hmm. That's odd. I think I tried but I got some error but now it seems to work.
Ok then I'm definitely putting this into an extension in test target

Tests/Helpers/TestHelpers.swift
imaginaris: Yes. Putting 0 was a very big mistake which caused weird race conditions that sometimes made PublicApiSpec test fail.
(because ping calls were spawned in thousands)

RInAppMessaging/Classes/Models/Responses/DisplayPermissionResponse.swift
imaginaris: Ah this init is used by another class in the SDK after all. I'm leaving it then. I'll remove the comment that it's for unit test only

RInAppMessaging/Classes/Models/Responses/DisplayPermissionResponse.swift
```diff
@@ -1,4 +1,33 @@
...
+
+    init(from decoder: Decoder) throws {
```
donnie-jp, line 19: If the below function calls can throw, should we be wrapping these in a catch?

RInAppMessaging/Classes/Models/Responses/DisplayPermissionResponse.swift
```diff
@@ -1,4 +1,33 @@
...
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        display = try container.decode(Bool.self, forKey: .display)
+        performPing = try container.decode(Bool.self, forKey: .performPing)
+        creationTimeMilliseconds = (try? container.decode(Int64.self, forKey: .creationTimeMilliseconds)) ?? Date().millisecondsSince1970
+    }
+
+    func encode(to encoder: Encoder) throws {
```
donnie-jp, line 26: If the below function calls can throw, should we be wrapping these in a catch?

RInAppMessaging/Classes/Protocols/TaskSchedulable.swift
```diff
@@ -8,16 +8,16 @@ internal protocol TaskSchedulable: AnyObject {
...
     /// - Parameter task: A block of code to run after the delay.
     /// - Parameter wallDeadline: Set `true` to use wall time (gettimeofday) instead of absolute time to ensure execution after returning from background
-    func scheduleWorkItem(milliseconds: Int, wallDeadline: Bool, closure: @escaping () -> Void)
+    func scheduleTask(milliseconds: Int, wallDeadline: Bool, _ task: @escaping () -> Void)
 }
 
 /// Default implementation.
 extension TaskSchedulable {
-    func scheduleWorkItem(milliseconds: Int, wallDeadline: Bool, closure: @escaping () -> Void) {
+    func scheduleTask(milliseconds: Int, wallDeadline: Bool, _ task: @escaping () -> Void) {
```
donnie-jp, line 16: Is this change so it can be called as a trailing closure?

RInAppMessaging/Classes/Models/Responses/DisplayPermissionResponse.swift
imaginaris: no. The responsibility is passed to the caller of this init :) which is marked as `throws`

RInAppMessaging/Classes/Protocols/TaskSchedulable.swift
imaginaris: No, it's just about naming unification

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
```diff
@@ -70,29 +89,50 @@ internal class CampaignRepository: CampaignRepositoryType {
...
         guard let index = list.firstIndex(where: { $0.id == campaign.id }) else {
-            Logger.debug("Campaign \(campaign.id) cannot be updated - not found in repository")
-            return false
+            Logger.debug("Campaign \(campaign.id) could not be updated - not found in the repository")
+            assertionFailure()
+            return nil
         }
-        list[index] = campaign
+
+        let udpatedCampaign = Campaign.updatedCampaign(campaign, withImpressionLeft: newValue)
```
donnie-jp: udpatedCampaign -> updatedCampaign

RInAppMessaging/Classes/UserDataCache.swift
```diff
@@ -0,0 +1,77 @@
...
+
+    func displayPermissionData(for campaign: Campaign) -> DisplayPermissionResponse? {
+        return displayPermissionData[campaign.id]
+    }
+}
+
+internal class UserDataCache: UserDataCacheable {
+
+    private let userDefaults: UserDefaults
+    private var cachedContainers: [Set<UserIdentifier>: UserDataCacheContainer]
```
donnie-jp: Would it be helpful to alias that type for conciseness?

RInAppMessaging/Classes/UserDataCache.swift
```diff
@@ -0,0 +1,77 @@
...
+        self.userDefaults = userDefaults
+
+        if let persistedData = userDefaults.object(forKey: persistedDataKey) as? Data {
+            do {
+                let decodedData = try JSONDecoder().decode([Set<UserIdentifier>: UserDataCacheContainer].self, from: persistedData)
+                cachedContainers = decodedData
+            } catch {
+                cachedContainers = [:]
+                Logger.debug("UserDataCache decoding failed! \(error)")
+                assertionFailure()
```
donnie-jp: This scenario isn't covered by the tests according to CodeCov.

RInAppMessaging/Classes/UserDataCache.swift
```diff
@@ -0,0 +1,77 @@
...
+        var currentData = cachedContainers[cacheKey] ?? UserDataCacheContainer()
+        currentData.campaignData = data
+        cachedContainers[cacheKey] = currentData
+        saveData()
+    }
+
+    func cacheDisplayPermissionData(_ data: DisplayPermissionResponse, for campaign: Campaign, userIdentifiers: [UserIdentifier]) {
+        let cacheKey = userKey(from: userIdentifiers)
+        var currentData = cachedContainers[cacheKey] ?? UserDataCacheContainer()
+        currentData.displayPermissionData[campaign.id] = data
```
donnie-jp: If you just need the id, should that be the param rather than campaign?

RInAppMessaging/Classes/UserDataCache.swift
imaginaris: Yeah it might be 👍

RInAppMessaging/Classes/UserDataCache.swift
donnie-jp: I'm not that clear on how you're constructing the key here.

Also, since it is made up of user identifiers possibly including usernames I think we should salt + hash as well. What do you think?

RInAppMessaging/Classes/UserDataCache.swift
imaginaris: Ah that's right. I should write some tests for this 👍

malayli (approved):
LGTM. I think that there are still some comments from @donnie-jp

RInAppMessaging/Classes/UserDataCache.swift
imaginaris: Ah good one. Yes we need to have some kind of encryption here

RInAppMessaging/Classes/Models/Responses/DisplayPermissionResponse.swift
donnie-jp: ok 🙂 

and is the caller the `try JSONDecoder().decode(DisplayPermissionResponse.self` that's wrapped in a do-catch?

RInAppMessaging/Classes/UserDataCache.swift
imaginaris: I'll add it in the next PR

RInAppMessaging/Classes/Models/Responses/DisplayPermissionResponse.swift
imaginaris: Yes 🙂

RInAppMessaging/Classes/UserDataCache.swift
imaginaris: Fixed

RInAppMessaging/Classes/UserDataCache.swift
donnie-jp: Created SDKCF-3667 so we don't forget 😄

donnie-jp (approved):
LGTM
Great work 👍

PR #57, opened 2021-04-18
SDKCF-3652: Get config new endpoint

malayli:
# Description
- Change Config endpoint call to be a GET with query params instead of: POST + JSON body.
- Add rolloutPercentage property to the ConfigData model. 

## Links
SDKCF-3652

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/Models/Responses/GetConfigResponse.swift
+++ b/RInAppMessaging/Classes/Models/Responses/GetConfigResponse.swift
@@ -3,8 +3,15 @@ internal struct GetConfigResponse: Decodable {
 }
 
 internal struct ConfigData: Decodable {
-    let enabled: Bool
+    let rolloutPercentage: Int
     let endpoints: EndpointURL?
+}
+
+extension ConfigData {
+    // Temporary property - It has to be removed in the ticket SDKCF-3663
+    var enabled: Bool {
+        rolloutPercentage > 0
+    }
 }
 
 internal struct EndpointURL: Decodable, Equatable {
--- a/RInAppMessaging/Classes/Protocols/HttpRequestable.swift
+++ b/RInAppMessaging/Classes/Protocols/HttpRequestable.swift
@@ -13,6 +13,9 @@ internal enum RequestError: Error {
     case taskFailed(Error)
     case httpError(Int, URLResponse?, Data?)
     case bodyEncodingError(Error?)
+    case urlBuildingError(Error)
+    case urlIsNil
+    case bodyIsNil
 }
 
 internal protocol HttpRequestable {
@@ -32,7 +35,15 @@ internal protocol HttpRequestable {
                            addtionalHeaders: [HeaderAttribute]?,
                            completion: @escaping (_ result: RequestResult) -> Void)
 
+    func buildURLRequest(url: URL) -> Result<URLRequest, Error>
+
     func buildHttpBody(with parameters: [String: Any]?) -> Result<Data, Error>
+}
+
+extension HttpRequestable {
+    func buildURLRequest(url: URL) -> Result<URLRequest, Error> {
+        .success(URLRequest(url: url))
+    }
 }
 
 /// Default implementation of HttpRequestable.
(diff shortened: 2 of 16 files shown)


RInAppMessaging/Classes/Models/Responses/GetConfigResponse.swift
```diff
@@ -3,6 +3,7 @@ internal struct GetConfigResponse: Decodable {
 }
 
 internal struct ConfigData: Decodable {
+    let rolloutPercentage: Int
     let enabled: Bool
```
malayli: `enabled` will be removed in an other ticket.

Tests/ConfigurationServiceSpec.swift
```diff
@@ -33,7 +33,8 @@ class ConfigurationServiceSpec: QuickSpec {
                         done()
                     }
                 }
-                expect(httpSession.sentRequest?.url).to(equal(configURL))
+                expect(httpSession.sentRequest?.url?.scheme).to(equal(configURL.scheme))
+                expect(httpSession.sentRequest?.url?.host).to(equal(configURL.host))
```
malayli, line 37: `configURL` has query items as it is now a GET request.
So `scheme` (http) and `host` (config.url) are used to match the previous test (http://config.url).

RInAppMessaging/Classes/Protocols/HttpRequestable.swift
```diff
@@ -32,9 +35,17 @@ internal protocol HttpRequestable {
...
                            completion: @escaping (_ result: RequestResult) -> Void)
 
+    func buildURLRequest(url: URL) -> Result<URLRequest, Error>
+
     func buildHttpBody(with parameters: [String: Any]?) -> Result<Data, Error>
 }
 
+extension HttpRequestable {
+    func buildURLRequest(url: URL) -> Result<URLRequest, Error> {
+        .success(URLRequest(url: url))
```
donnie-jp, line 45: Is this always returning success? It can't fail?

Tests/ConfigurationRepositorySpec.swift
```diff
@@ -27,15 +27,15 @@ class ConfigurationRepositorySpec: QuickSpec {
                 let endpoints = EndpointURL(ping: URL(string: "ping")!,
                                             displayPermission: URL(string: "displayPermission")!,
                                             impression: URL(string: "impression")!)
-                let config = ConfigData(enabled: true, endpoints: endpoints)
+                let config = ConfigData(rolloutPercentage: 45, enabled: true, endpoints: endpoints)
```
donnie-jp: You could change the value 45 into a constant defined in the tests.

RInAppMessaging/Classes/Models/Responses/GetConfigResponse.swift
donnie-jp: OK

donnie-jp (approved):
A few comments but LGTM 👍

RInAppMessaging/Classes/Protocols/HttpRequestable.swift
malayli: It can't fail. It is the default implementation of `buildURLRequest`.

RInAppMessaging/Classes/Services/ConfigurationService.swift
```diff
@@ -54,31 +54,39 @@ internal struct ConfigurationService: ConfigurationServiceType, HttpRequestable
...
 
+    func buildURLRequest(url: URL) -> Result<URLRequest, Error> {
         do {
-            let body = try JSONEncoder().encode(getConfigRequest)
-            return .success(body)
+            let request = try getConfigRequest()
+            var urlComponents = URLComponents(url: url, resolvingAgainstBaseURL: true)
+            urlComponents?.queryItems = request.toQueryItems
+            guard let url = urlComponents?.url else {
+                return .failure(RequestError.urlIsNil)
```
imaginaris, line 79: Isn't it better to use existing `urlBuildingError`?

malayli: urlBuildingError takes an error as a parameter.

imaginaris: Ah ok I can see it later wrapped in it. All fine then

imaginaris (approved):
LGTM

PR #59, opened 2021-04-19
improve: hash the key used to access cache (SDKCF-3667)

imaginaris:
# Description
This hashing method was easiest to implement.
We can discuss alternative approach if this one is not secure enough.

## Links
SDKCF-3667

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/UserDataCache.swift
+++ b/RInAppMessaging/Classes/UserDataCache.swift
@@ -79,9 +79,9 @@ internal class UserDataCache: UserDataCacheable {
         identifiers.map({ $0.identifier }).sorted().forEach {
             hasher.combine($0)
         }
-        hasher.encryprtionMethod = .md5
+        hasher.encryptionMethod = .md5
         let salt = hasher.generateHash()
-        hasher.encryprtionMethod = .sha256
+        hasher.encryptionMethod = .sha256
         hasher.salt = salt
 
         return hasher.generateHash()
--- a/RInAppMessaging/Classes/UserDataCache.swift
+++ b/RInAppMessaging/Classes/UserDataCache.swift
@@ -20,7 +20,7 @@ internal struct UserDataCacheContainer: Codable, Equatable {
 
 internal class UserDataCache: UserDataCacheable {
 
-    private typealias CacheContainers = [Set<UserIdentifier>: UserDataCacheContainer]
+    private typealias CacheContainers = [String: UserDataCacheContainer]
 
     private let userDefaults: UserDefaults
     private var cachedContainers: CacheContainers
@@ -74,7 +74,16 @@ internal class UserDataCache: UserDataCacheable {
         }
     }
 
-    private func userKey(from identifiers: [UserIdentifier]) -> Set<UserIdentifier> {
-        Set<UserIdentifier>(identifiers)
+    private func userKey(from identifiers: [UserIdentifier]) -> String {
+        var hasher = KeyHasher()
+        identifiers.map({ $0.identifier }).sorted().forEach {
+            hasher.combine($0)
+        }
+        hasher.encryprtionMethod = .md5
+        let salt = hasher.generateHash()
+        hasher.encryprtionMethod = .sha256
+        hasher.salt = salt
+
+        return hasher.generateHash()
     }
 }
--- a/RInAppMessaging/Classes/Utilities/KeyHasher.swift
+++ b/RInAppMessaging/Classes/Utilities/KeyHasher.swift
@@ -1,0 +1,79 @@
+import CommonCrypto
+import CryptoKit
+
+/// Tool for creating a salted hash from given string values using selected KeyHasher.Encryption method
+internal struct KeyHasher {
+
+    enum Encryption: String {
+        case md5
+        case sha256
+    }
+
+    var encryprtionMethod = Encryption.sha256
+    var salt: String?
+
+    private var data = [String]()
+
+    mutating func combine(_ value: String) {
+        data.append(value)
+    }
+
+    mutating func generateHash() -> String {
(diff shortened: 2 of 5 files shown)


RInAppMessaging/Classes/UserDataCache.swift
```diff
@@ -74,7 +74,16 @@ internal class UserDataCache: UserDataCacheable {
...
 
-    private func userKey(from identifiers: [UserIdentifier]) -> Set<UserIdentifier> {
-        Set<UserIdentifier>(identifiers)
+    private func userKey(from identifiers: [UserIdentifier]) -> String {
+        var hasher = KeyHasher()
+        identifiers.map({ $0.identifier }).sorted().forEach {
+            hasher.combine($0)
+        }
+        hasher.encryprtionMethod = .md5
+        let salt = hasher.generateHash()
```
donnie-jp, line 83: So the salt is the md5 hash of the user ids?

imaginaris: Yes. I'm not sure if it makes key more secure but it's something 🙂

malayli (approved):

RInAppMessaging/Classes/Utilities/KeyHasher.swift
```diff
@@ -0,0 +1,79 @@
...
+                _ = CC_MD5($0.baseAddress, CC_LONG(data.count), &hash)
+            }
+            return Data(hash).hexString
+        }
+    }
+
+    private func sha256(string: String) -> String {
+        let stringData = Data(string.utf8)
+
+        if #available(iOS 13.0, *) {
```
donnie-jp: We always need to compute a hash regardless of the OS version so it might be better to use the CommonCrypto pre-iOS 13 approach for all iOS versions and remove CryptoKit. That way we know the hashing is always getting run/tested.

What do you think @imaginaris / @malayli ?

Tests/KeyHasherSpec.swift
```diff
@@ -0,0 +1,126 @@
+import Quick
+import Nimble
+@testable import RInAppMessaging
+
+class KeyHasherSpec: QuickSpec {
+
+    override func spec() {
+        // Be sure to run those tests for ios version >=13.0 and <13.0
```
donnie-jp: See my comment above about CryptoKit / CommonCrypto hashing.

Tests/KeyHasherSpec.swift
```diff
@@ -0,0 +1,126 @@
...
+
+                    beforeEach {
+                        hasher.encryprtionMethod = encryption
+                    }
+
+                    it("will generate a valid hash for empty data") {
+                        expect(hasher.generateHash()).toNot(beEmpty())
+                    }
+
+                    it("will generate a valid hash for empty value") {
```
donnie-jp, line 28: Is this the anonymous (no user ids) case?

RInAppMessaging/Classes/Utilities/KeyHasher.swift
```diff
@@ -0,0 +1,79 @@
...
+
+/// Tool for creating a salted hash from given string values using selected KeyHasher.Encryption method
+internal struct KeyHasher {
+
+    enum Encryption: String {
+        case md5
+        case sha256
+    }
+
+    var encryprtionMethod = Encryption.sha256
```
donnie-jp: encryprtionMethod -> encryptionMethod

donnie-jp (changes requested):
Pretty much LGTM but I want to resolve a couple of discussion points before approving.

RInAppMessaging/Classes/Utilities/KeyHasher.swift
imaginaris: That makes sense. I thought that it's always better to use new features if possible but in case of core logic or security it should be more about consistence and stability. So I think you are right, we should keep only pre-iOS 13 approach

Tests/KeyHasherSpec.swift
imaginaris: hmm I think it's rather the "no data" one.

RInAppMessaging/Classes/Utilities/KeyHasher.swift
imaginaris: I like AndroidStudio more because it checks for typos...

RInAppMessaging/Classes/Utilities/KeyHasher.swift
```diff
@@ -37,38 +36,21 @@ internal struct KeyHasher {
...
-            return hash.hexString
-        } else {
-            var hash = [UInt8](repeating: 0, count: Int(CC_MD5_DIGEST_LENGTH))
-            data.withUnsafeBytes {
-                _ = CC_MD5($0.baseAddress, CC_LONG(data.count), &hash)
-            }
-            return Data(hash).hexString
+        var hash = [UInt8](repeating: 0, count: Int(CC_MD5_DIGEST_LENGTH))
+        stringData.withUnsafeBytes {
+            _ = CC_MD5($0.baseAddress, CC_LONG(stringData.count), &hash)
```
donnie-jp, line 41: Do we need to check in case the stringData.count length is > CC_{ALG}_DIGEST_LENGTH (for MD5 and SHA256 algs) to avoid a possible buffer overrun crash?

imaginaris: Hmm I don't think so. You mean when string is longer than hash length?  That should be fine (I added a test for that)

donnie-jp: Yes, I meant if string is longer than hash buffer. 
OK, I guess CommonCrypto must have a check internally then.

donnie-jp (approved):
LGTM

PR #60, opened 2021-04-20
SDKCF-3707: Integration tests fixes

malayli:
# Description
Integration tests fixes.

## Links
SDKCF-3707

# Checklist

### DIFF
--- a/IntegrationTests/IntegrationTests.swift
+++ b/IntegrationTests/IntegrationTests.swift
@@ -42,10 +42,10 @@ class IntegrationTests: XCTestCase {
                 response = nil
             }
             XCTAssertNotNil(response)
-            XCTAssert(response?.enabled == true)
-            XCTAssertNotNil(response?.endpoints.ping)
-            XCTAssertNotNil(response?.endpoints.displayPermission)
-            XCTAssertNotNil(response?.endpoints.impression)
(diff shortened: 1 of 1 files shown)


IntegrationTests/IntegrationTests.swift
```diff
@@ -42,10 +42,10 @@ class IntegrationTests: XCTestCase {
                 response = nil
             }
             XCTAssertNotNil(response)
-            XCTAssert(response?.enabled == true)
-            XCTAssertNotNil(response?.endpoints.ping)
-            XCTAssertNotNil(response?.endpoints.displayPermission)
-            XCTAssertNotNil(response?.endpoints.impression)
+            XCTAssert(response?.rolloutPercentage == 100)
```
imaginaris: Can we assume that rollout will always be 100? Maybe just check if it's not nil or >0?

malayli: Checking not nil is better I think.

imaginaris: Ah right it can be 0. I think checking for nil is enough

imaginaris (approved):

donnie-jp:
Thanks for fixing these @malayli

PR #62, opened 2021-04-21
improve: handle transition of anonymous user -> logged-in user (SDKCF-3710)

imaginaris:
# Description
- Added syncing of anonymous cache and logged-in user cache.
(anon cache is used before logged-in user data reaches the SDK)
- Fixed impression counter update when campaign display is cancelled (onVerify() and closeMessage() API)

## Links
SDKCF-3710

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/InAppMessagingModule.swift
+++ b/RInAppMessaging/Classes/InAppMessagingModule.swift
@@ -92,15 +92,17 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
         } else {
             diff = preferenceRepository.preference == nil ? [] : nil
         }
+        let isFirstUser = preferenceRepository.getUserIdentifiers().isEmpty && diff?.isEmpty == false
         preferenceRepository.setPreference(preference)
 
         guard isInitialized else {
             return
         }
         if diff?.isEmpty != true && diff != [.accessToken] {
+            // reset only if identifiers changed
             readyCampaignDispatcher.resetQueue()
         }
-        campaignRepository.loadCachedData()
+        campaignRepository.loadCachedData(syncWithDefaultUserData: isFirstUser)
         campaignsListManager.refreshList()
     }
 
--- a/RInAppMessaging/Classes/Repositories/CampaignRepository.swift
+++ b/RInAppMessaging/Classes/Repositories/CampaignRepository.swift
@@ -28,7 +28,7 @@ internal protocol CampaignRepositoryType: AnyObject, Lockable {
     func incrementImpressionsLeftInCampaign(id: String) -> Campaign?
 
     /// Loads campaign data from user cache
-    func loadCachedData()
+    func loadCachedData(syncWithDefaultUserData: Bool)
 }
 
 /// Repository to store campaigns retrieved from ping request.
@@ -38,6 +38,8 @@ internal class CampaignRepository: CampaignRepositoryType {
     private let preferenceRepository: IAMPreferenceRepository
     private let campaigns = LockableObject([Campaign]())
     private(set) var lastSyncInMilliseconds: Int64?
+    private var defaultCacheUser = [UserIdentifier]()
+
     var list: [Campaign] {
         return campaigns.get()
     }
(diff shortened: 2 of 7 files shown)


RInAppMessaging/Classes/Repositories/CampaignRepository.swift
```diff
@@ -109,11 +112,22 @@ internal class CampaignRepository: CampaignRepositoryType {
         return updateImpressionsLeftInCampaign(campaign, newValue: campaign.impressionsLeft + 1)
     }
 
-    func loadCachedData() {
-        let cachedData = userDataCache.getUserData(identifiers: preferenceRepository.getUserIdentifiers())?.campaignData
-        campaigns.set(value: cachedData ?? [])
+    func loadCachedData(syncWithDefaultUserData: Bool) {
+        var cachedData = userDataCache.getUserData(identifiers: preferenceRepository.getUserIdentifiers())?.campaignData ?? []
```
donnie-jp, line 116: Is this the cached data for the newly logged in user?

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
```diff
@@ -109,11 +112,22 @@ internal class CampaignRepository: CampaignRepositoryType {
...
+        if syncWithDefaultUserData {
+            userDataCache.getUserData(identifiers: defaultCacheUser)?.campaignData?.forEach({ defaultUserCampaign in
+                if let existingCampaignIndex = cachedData.firstIndex(where: { $0.id == defaultUserCampaign.id }) {
+                    cachedData[existingCampaignIndex] = defaultUserCampaign
```
donnie-jp, line 120: So where there's already an existing campaign in the default user's cached data we replace the newly logged in user's campaign with that existing campaign?

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
```diff
@@ -129,7 +143,16 @@ internal class CampaignRepository: CampaignRepositoryType {
...
+        saveDataToCache(list)
+
         return updatedCampaign
     }
+
+    private func saveDataToCache(_ list: [Campaign]) {
+        let user = preferenceRepository.getUserIdentifiers()
+        userDataCache.cacheCampaignData(list, userIdentifiers: user)
+        if defaultCacheUser != user {
+            userDataCache.cacheCampaignData(list, userIdentifiers: defaultCacheUser)
```
donnie-jp, line 155: Why do we need this?

RInAppMessaging/Classes/InAppMessagingModule.swift
```diff
@@ -92,15 +92,17 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
         } else {
             diff = preferenceRepository.preference == nil ? [] : nil
         }
+        let isFirstUser = preferenceRepository.getUserIdentifiers().isEmpty && diff?.isEmpty == false
```
donnie-jp, line 95: Does this mean the first logged in user? i.e. non-default user

donnie-jp (approved):
Basically LGTM - very thorough unit tests 👍 

I left some comments that I'd like you to respond to but it doesn't block merging.

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
imaginaris: Might be. Or it can be also anonymous user cache. Depends on what's returned in `getUserIdentifiers()`

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
imaginaris: Yes. I hope that logic doesn't have any holes

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
imaginaris: Any update in logged-in user cache is also reflected in anon/default user cache so the same data can be picked up at the next app launch. Keep in mind that there's ping call in between which additionally ensures that campaign list is up to date

RInAppMessaging/Classes/InAppMessagingModule.swift
imaginaris: Yes. I should probably use a better name for this var

PR #63, opened 2021-04-22
SDKCF-3646: logEvent crash fix - shared queue

malayli:
# Description
Originally I added the queue to the container to easily set the resumeQueue to ConfigurationManager.
The crash occurs when the queue is resolved in logEvent method.
It is not the best fix but it should solve this crash.

We can decline this PR if we really want to fix the resolving issue in DependencyManager.

Fixes:
- It is safer to use a static immutable queue without resolving it each time we want to use it. It prevents to have a invalid reference.
- It solves the situation when dependencyManager is deallocated by IAM.Main thread because of config response telling SDK to disable, and at the same time dependencyManager is accessed from main thread during logEvent() to get the queue reference.

## Links
SDKCF-3646

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/DependencyManager/MainContainer.swift
+++ b/RInAppMessaging/Classes/DependencyManager/MainContainer.swift
@@ -33,14 +33,11 @@ internal enum MainContainerFactory {
             ContainerElement(type: ConfigurationRepositoryType.self, factory: {
                 ConfigurationRepository()
             }),
-            ContainerElement(type: DispatchQueue.self, factory: {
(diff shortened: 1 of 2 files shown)


malayli:
The 2 Security Hotspots are about md5 functions in KeyHasher.

imaginaris:
> The 2 Security Hotspots are about md5 functions in KeyHasher.

@donnie-jp Can you review those issues in Sonarcloud and change their status to make the PR errors go?

imaginaris (approved):

donnie-jp (approved):
LGTM

PR #64, opened 2021-04-22
improve: add subscription id in config api header SDKCF-3716

donnie-jp:
# Description
Add subscription id to Config API request header

## Links
SDKCF-3716

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/Services/ConfigurationService.swift
+++ b/RInAppMessaging/Classes/Services/ConfigurationService.swift
@@ -23,7 +23,8 @@ internal struct ConfigurationService: ConfigurationServiceType, HttpRequestable 
         let response = requestFromServerSync(
             url: configURL,
             httpMethod: .get,
-            addtionalHeaders: nil)
+            addtionalHeaders: buildRequestHeader()
+        )
 
         switch response {
         case .success((let data, _)):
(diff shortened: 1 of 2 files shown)


RInAppMessaging/Classes/Services/ConfigurationService.swift
```diff
@@ -23,7 +23,8 @@ internal struct ConfigurationService: ConfigurationServiceType, HttpRequestable
         let response = requestFromServerSync(
             url: configURL,
             httpMethod: .get,
-            addtionalHeaders: nil)
+            addtionalHeaders: buildRequestHeader()
```
donnie-jp, line 26: The implementation was copied/modified from MessageMixerService.

RInAppMessaging/Classes/Services/ConfigurationService.swift
```diff
@@ -70,7 +71,21 @@ extension ConfigurationService {
...
-    func buildURLRequest(url: URL) -> Result<URLRequest, Error> {
+    private func buildRequestHeader() -> [HeaderAttribute] {
+        let Keys = Constants.Request.Header.self
+        var additionalHeaders: [HeaderAttribute] = []
+
+        if let subId = bundleInfo.inAppSubscriptionId, !subId.isEmpty {
+            additionalHeaders.append(HeaderAttribute(key: Keys.subscriptionID, value: subId))
+        } else {
+            Logger.debug("Info.plist must contain a valid InAppMessagingAppSubscriptionID")
+            assertionFailure()
```
donnie-jp, line 82: It's safe to call this in production release code isn't it?

imaginaris: If we release with optimization setting `-O` (which we do) then that method has no effect. (So yes, it's safe 🙂 )

imaginaris (approved):

malayli (approved):

PR #65, opened 2021-04-22
fix: fix user data cache sync (SDKCF-3710)

imaginaris:
# Description
Anonymous user has now its own independent cache.
Last user cache is used only after app launch and before previously logged in user data is registered (and there was no logout  in previous app session)

## Links
SDKCF-3710

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -23,7 +23,7 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
 
     private let dispatchQueue = DispatchQueue(label: "IAM.Campaign", qos: .userInteractive)
     private var queuedCampaigns = [Campaign]()
-    @AtomicGetSet private(set) var isDispatching = false
+    private(set) var isDispatching = false
 
     weak var delegate: CampaignDispatcherDelegate?
     var scheduledTask: DispatchWorkItem?
@@ -44,9 +44,13 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
     }
 
     func resetQueue() {
-        isDispatching = false
-        scheduledTask?.cancel()
         dispatchQueue.async {
+            let isDisplayingCampaign = self.scheduledTask == nil && self.isDispatching
+            if !isDisplayingCampaign {
+                self.isDispatching = false
+                self.scheduledTask?.cancel()
+            }
+
             self.queuedCampaigns.removeAll()
         }
     }
--- a/RInAppMessaging/Classes/Protocols/Lockable.swift
+++ b/RInAppMessaging/Classes/Protocols/Lockable.swift
@@ -12,6 +12,9 @@ internal protocol LockableResource {
     func unlock()
 }
 
+// The following rule is disabled to ensure thread safety
+// swiftlint:disable shorthand_operator
+
 /// Object-wrapper that conforms to LockableResource protocol.
 /// Used to control getter and setter of given resource.
 /// When lock() has been called on some thread, only that thread will be able to access the resource.
@@ -42,15 +45,15 @@ internal class LockableObject<T>: LockableResource {
         if shouldWait {
             dispatchGroup.wait()
         }
-        lockCount += 1
+        lockCount = lockCount + 1
         lockingThread = Thread.current
         dispatchGroup.enter()
     }
 
     func unlock() {
         if lockCount > 0 {
+            lockCount = lockCount - 1
             dispatchGroup.leave()
-            lockCount -= 1
         } else {
             lockingThread = nil
         }
--- a/RInAppMessaging/Classes/RInAppMessaging.swift
+++ b/RInAppMessaging/Classes/RInAppMessaging.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 /// Protocol for optional delagate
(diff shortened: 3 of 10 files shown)


RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -44,9 +44,13 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
     }
 
     func resetQueue() {
-        isDispatching = false
-        scheduledTask?.cancel()
```
donnie-jp, line 48: Why were these moved into the async q?

RInAppMessaging/Classes/Protocols/Lockable.swift
```diff
@@ -42,15 +45,15 @@ internal class LockableObject<T>: LockableResource {
         if shouldWait {
             dispatchGroup.wait()
         }
-        lockCount += 1
+        lockCount = lockCount + 1
```
donnie-jp, line 48: How does this ensure thread safety?
Or is it because you've moved the decrement before the leave()?

RInAppMessaging/Classes/RInAppMessaging.swift
```diff
@@ -1,3 +1,5 @@
+import Foundation
```
donnie-jp, line 1: Is that import required?

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
```diff
@@ -28,17 +28,21 @@ internal protocol CampaignRepositoryType: AnyObject, Lockable {
...
+    func loadCachedData(syncWithLastUserData: Bool)
+
+    /// Deletes cached data used to sync between users
+    func resetDataPersistence()
 }
 
 /// Repository to store campaigns retrieved from ping request.
 internal class CampaignRepository: CampaignRepositoryType {
 
+    static let lastUser = [UserIdentifier(type: .userId, identifier: "IAM.lastUser!@#")]
```
donnie-jp, line 40: Is the !@# just for uniqueness?

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
```diff
@@ -151,8 +159,6 @@ internal class CampaignRepository: CampaignRepositoryType {
     private func saveDataToCache(_ list: [Campaign]) {
         let user = preferenceRepository.getUserIdentifiers()
         userDataCache.cacheCampaignData(list, userIdentifiers: user)
-        if defaultCacheUser != user {
-            userDataCache.cacheCampaignData(list, userIdentifiers: defaultCacheUser)
-        }
+        userDataCache.cacheCampaignData(list, userIdentifiers: CampaignRepository.lastUser)
```
donnie-jp, line 162: Why is the same campaign list cached twice?

donnie-jp (approved):
LGTM - I left a few questions asking for clarifications but we can proceed to merge this so it can be QA'd.

RInAppMessaging/Classes/UserDataCache.swift
```diff
@@ -2,6 +2,7 @@ internal protocol UserDataCacheable: AnyObject {
     func getUserData(identifiers: [UserIdentifier]) -> UserDataCacheContainer?
     func cacheCampaignData(_ data: [Campaign], userIdentifiers: [UserIdentifier])
     func cacheDisplayPermissionData(_ data: DisplayPermissionResponse, campaignID: String, userIdentifiers: [UserIdentifier])
+    func deleteUserData(identifiers: [UserIdentifier])
```
donnie-jp, line 5: Where is the separate anonymous user cache that was mentioned in the PR description?

RInAppMessaging/Classes/CampaignDispatcher.swift
imaginaris: Because I think it wasn't really necessary to keep those outside, and now it's more thread safe.

RInAppMessaging/Classes/Protocols/Lockable.swift
imaginaris: `@AtomicGetSet` doesn't work for incrementations. But actually this change doesn't help either. I'll have to fix that :)
Moving it before `leave()` might help though.

RInAppMessaging/Classes/RInAppMessaging.swift
imaginaris: Sometimes when I clean the project I get a warning/error for a moment that I'm using `Foundation` features (`@objc`) without importing it. It works without that import in our app but it might not for others.

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
imaginaris: Yes... 😬

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
imaginaris: It's basically the same approach as before but it doesn't store to anonymous user cache (to avoid userA -> logout -> userB data transfer)

RInAppMessaging/Classes/UserDataCache.swift
imaginaris: In CampaignRepository `lastUser`

PR #67, opened 2021-04-25
improve: add `mutate` function to `AtomicGetSet` property wrapper

imaginaris:
# Description
Added `mutate` method for struct properties wrapped in `@AtomicGetSet`.
Examples:
```swift
intVar.mutate { $0+=1 }
arrayVar.mutate { $0.append("newValue") }
dictVar.mutate { $0["key"] = "value" }
```

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
+++ b/RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
@@ -2,7 +2,7 @@
 /// Mutating functions, subscript, incrementation etc. are not synchronized by default -
 /// use `mutate` function to ensure operation atomicity.
 @propertyWrapper
-class AtomicGetSet<Value> {
+struct AtomicGetSet<Value> {
     private let queue = PropertyQueueGenerator.spawnQueue(domain: "IAM.AtomicProperty")
     private var value: Value
 
@@ -19,7 +19,7 @@ class AtomicGetSet<Value> {
         }
     }
 
-    func mutate(_ mutation: (inout Value) -> Void) {
+    mutating func mutate(_ mutation: (inout Value) -> Void) {
         queue.sync {
             mutation(&value)
         }
--- a/RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
+++ b/RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
@@ -1,7 +1,8 @@
 /// This wrapper ensures synchronized access to the value only for getter and setter.
-/// Mutating functions, subscript, incrementation etc. are not synchronized.
+/// Mutating functions, subscript, incrementation etc. are not synchronized by default -
+/// use `mutate` function to ensure operation atomicity.
 @propertyWrapper
-struct AtomicGetSet<Value> {
+class AtomicGetSet<Value> {
     private let queue = PropertyQueueGenerator.spawnQueue(domain: "IAM.AtomicProperty")
     private var value: Value
 
@@ -15,6 +16,12 @@ struct AtomicGetSet<Value> {
         }
         set {
             queue.sync { value = newValue }
+        }
+    }
+
+    func mutate(_ mutation: (inout Value) -> Void) {
+        queue.sync {
+            mutation(&value)
         }
     }
 }
--- a/Tests/AtomicWrapperSpec.swift
+++ b/Tests/AtomicWrapperSpec.swift
@@ -77,6 +77,27 @@ class AtomicWrapperSpec: QuickSpec {
                 dispatchGroup.wait()
             }
 
+            it("will not crash when one thread writes and the other reads the same value at the same time") {
+                let dispatchGroup = DispatchGroup()
+                dispatchGroup.enter()
+                dispatchGroup.enter()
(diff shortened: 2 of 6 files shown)


RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
```diff
@@ -1,7 +1,8 @@
 /// This wrapper ensures synchronized access to the value only for getter and setter.
-/// Mutating functions, subscript, incrementation etc. are not synchronized.
+/// Mutating functions, subscript, incrementation etc. are not synchronized by default -
+/// use `mutate` function to ensure operation atomicity.
 @propertyWrapper
-struct AtomicGetSet<Value> {
+class AtomicGetSet<Value> {
```
donnie-jp: Not possible to do the mutate as a struct?

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
```diff
@@ -17,6 +18,12 @@ struct AtomicGetSet<Value> {
             queue.sync { value = newValue }
         }
     }
+
+    func mutate(_ mutation: (inout Value) -> Void) {
+        queue.sync {
+            mutation(&value)
+        }
+    }
```
donnie-jp, line 26: This needs a test.
Can we use a unit test to verify its thread safety?

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
```diff
@@ -17,6 +18,12 @@ struct AtomicGetSet<Value> {
             queue.sync { value = newValue }
         }
     }
+
+    func mutate(_ mutation: (inout Value) -> Void) {
```
malayli: Can this function be private?

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
malayli: Do we intend to inherit from `AtomicGetSet`? If not, can it be `final class`?

imaginaris: Hmm I got some errors but it's seems that it's working with `struct` after all. I was also wondering if shadow copying the property because of `inout` in mutate() is going to cause any problems but I think it's ok because everything is synced

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
imaginaris: I didn't have any idea for a test but now I've been able to come up with something.

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
imaginaris: No. It's supposed to be used by callers

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
imaginaris: I'm setting it back to `struct`

Tests/AtomicWrapperSpec.swift
```diff
@@ -0,0 +1,85 @@
...
+            let queueA = DispatchQueue(label: "QueueA")
+            let queueB = DispatchQueue(label: "QueueB")
+
+            beforeEach {
+                self.atomicArray = []
+            }
+
+            context("when using mutating functions") {
+
+                it("should ensure atomicity when using `mutate` function") {
```
donnie-jp: Can you add a comment that explains this test please

Tests/AtomicWrapperSpec.swift
```diff
@@ -0,0 +1,85 @@
...
+                            dispatchGroup.leave()
+                        }
+                        dispatchGroup.wait()
+                    }
+
+                    let expected = [[String]](repeating: ["string 1", "string 2"], count: 100).flatMap({ $0 })
+                    expect(self.atomicArray).to(elementsEqual(expected))
+                }
+
+                it("should not expect atomic operation without using `mutate` function") {
```
donnie-jp: Can you add a comment that explains this test please

donnie-jp (changes requested):
LTGM

Before merging I want devs to discuss whether this should be included in PP and what QA would be needed for us to be confident at including this.

malayli (approved):

donnie-jp (approved):
👍

Tests/AtomicWrapperSpec.swift
```diff
@@ -0,0 +1,185 @@
...
+                }
+                queueB.async {
+                    for _ in (1...1_000_000) {
+                        _ = self.atomicArray
+                    }
+                    dispatchGroup.leave()
+                }
+                dispatchGroup.wait()
+            }
+
```
donnie-jp, line 57: Should there also be a reading/writing thread safety test?

imaginaris: Like one thread reads and the other one writes the value? Yeah I can add that.

PR #71, opened 2021-04-28
fix: clear event list on user change to prevent unexpected campaign displays (SDKCF-3727)

imaginaris:
# Description
This fixes a situation when userA logs multiple events for given campaign that has a matching trigger and the display limit was reached, so when userB logs in with fresh impression counter, all stacked events from previous user make the campaign display immediately (even multiple times until userB's impression limit is reached)

## Links
SDKCF-3727

# Checklist

### DIFF
--- a/Tests/EventMatcherSpec.swift
+++ b/Tests/EventMatcherSpec.swift
@@ -248,6 +248,27 @@ class EventMatcherSpec: QuickSpec {
                     expect(eventMatcher.containsAllMatchedEvents(for: campaign)).to(beFalse())
                 }
             }
+
+            context("when calling clearNonPersistentEvents") {
+
+                it("will clear all matched non-persistent events") {
+                    campaignRepository.list = [testCampaign]
+                    eventMatcher.matchAndStore(event: LoginSuccessfulEvent())
+                    expect(eventMatcher.matchedEvents(for: testCampaign)).toNot(beEmpty())
+
+                    eventMatcher.clearNonPersistentEvents()
+                    expect(eventMatcher.matchedEvents(for: testCampaign)).to(beEmpty())
+                }
+
+                it("will not clear persistent events") {
+                    campaignRepository.list = [testCampaign]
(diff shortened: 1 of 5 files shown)


Tests/EventMatcherSpec.swift
```diff
@@ -248,6 +248,27 @@ class EventMatcherSpec: QuickSpec {
                     expect(eventMatcher.containsAllMatchedEvents(for: campaign)).to(beFalse())
                 }
             }
+
+            context("when calling wasClearNonPersistentEvents") {
```
donnie-jp: Typo?
wasClearNonPersistentEvents -> clearNonPersistentEvents

imaginaris: Copy/paste leftover :/

Tests/InAppMessagingModuleSpec.swift
```diff
@@ -278,6 +278,29 @@ class InAppMessagingModuleSpec: QuickSpec {
...
+                                        eventMatcher.wasClearNonPersistentEventsCalled = false // reset
+                                        iamModule.registerPreference(prefB)
+                                        expect(eventMatcher.wasClearNonPersistentEventsCalled).to(beTrue())
+                                    }
+                            }
+
+                            it("will not clear event list when user did not log out or change to another user") {
+                                [(nil, aUser), (IAMPreference(), aUser),
+                                 (nil, nil), (nil, IAMPreference()),
+                                 (IAMPreference(), nil), (IAMPreference(), IAMPreference())]
```
donnie-jp, line 296: Why is (IAMPreference(), nil) not considered log out? Cos it's an empty prefs object?

donnie-jp (approved):
LGTM

Tests/InAppMessagingModuleSpec.swift
imaginaris: Yeah. prefs can be `nil` empty with nil fields (IAMPreference()) and empty with empty fields (which I didn't test)... It all depends on how host app is using the SDK.

PR #76, opened 2021-05-12
fix: fix a situation when incrementImpressionsLeft was called twice when calling `closeMessage` API

imaginaris:
# Description
`closeMessage` sends onDismiss call with cancelled flag and that already makes completion handler in Router to call `incrementImpressionsLeftInCampaign`

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/Router.swift
+++ b/RInAppMessaging/Classes/Router.swift
@@ -14,8 +14,7 @@ internal protocol RouterType: AnyObject {
                          completion: @escaping (_ cancelled: Bool) -> Void)
 
     /// Removes displayed campaign view from the stack
-    /// - Returns: a campaign associated with displayed message or nil if no message was displayed
-    func discardDisplayedCampaign() -> Campaign?
+    func discardDisplayedCampaign()
 }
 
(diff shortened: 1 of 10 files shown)


RInAppMessaging/Classes/Router.swift
```diff
@@ -29,21 +28,17 @@ internal class Router: RouterType {
...
     }
 
-    func discardDisplayedCampaign() -> Campaign? {
-        return displayQueue.sync {
-            let discard: () -> Campaign? = {
-                guard let rootView = UIApplication.shared.getKeyWindow() else {
-                    return nil
+    func discardDisplayedCampaign() {
+        displayQueue.sync {
+            DispatchQueue.main.async {
```
donnie-jp, line 33: Is the main q needed because you're calling `UIApplication` API?

donnie-jp (approved):
LGTM

RInAppMessaging/Classes/Router.swift
imaginaris: Yes. especially the `removeFromSuperview()` part

PR #78, opened 2021-05-19
improve: defer events that were logged before SDK was initialized (SDKCF-3821)

imaginaris:
# Description
This improvement ensures the desired order of API calls being executed when getConfig request is processing or was scheduled (retry logic).

## Links
SDKCF-3821

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/InAppMessagingModule.swift
+++ b/RInAppMessaging/Classes/InAppMessagingModule.swift
@@ -121,12 +121,12 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
 
     private func flushEventBuffer(void: Bool) {
         if !void {
-            self.eventBuffer.forEach {
-                self.eventMatcher.matchAndStore(event: $0)
+            eventBuffer.forEach {
+                eventMatcher.matchAndStore(event: $0)
             }
             campaignTriggerAgent.validateAndTriggerCampaigns()
         }
-        self.eventBuffer.removeAll()
+        eventBuffer.removeAll()
     }
 }
 
--- a/RInAppMessaging/Classes/InAppMessagingModule.swift
+++ b/RInAppMessaging/Classes/InAppMessagingModule.swift
@@ -64,9 +64,9 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
 
             if enabled {
                 self.campaignsListManager.refreshList()
-                self.flushEventBuffer(void: false)
+                self.flushEventBuffer(discardEvents: false)
             } else {
-                self.flushEventBuffer(void: true)
+                self.flushEventBuffer(discardEvents: true)
                 deinitHandler()
             }
         }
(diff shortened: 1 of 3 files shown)


RInAppMessaging/Classes/InAppMessagingModule.swift
```diff
@@ -109,6 +118,16 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
         }
         router.discardDisplayedCampaign()
     }
+
+    private func flushEventBuffer(void: Bool) {
+        if !void {
+            self.eventBuffer.forEach {
```
donnie-jp: Is `self.` necessary in this function?

imaginaris: not necessary. Copy&paste issue. The same goes for other `self.` in this method

RInAppMessaging/Classes/InAppMessagingModule.swift
```diff
@@ -109,6 +118,16 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
         }
         router.discardDisplayedCampaign()
     }
+
+    private func flushEventBuffer(void: Bool) {
```
donnie-jp: Not sure about naming the param `void`.

Maybe `discardEvents` is clearer?

imaginaris: Me neither 😞 Any ideas? 'dispose'?

donnie-jp (approved):
LGTM - made a couple of small suggestions

RInAppMessaging/Classes/InAppMessagingModule.swift
donnie-jp: dispose, delete or discard (I updated my comment above) are clearer I think

Tests/InAppMessagingModuleSpec.swift
```diff
@@ -142,6 +158,62 @@ class InAppMessagingModuleSpec: QuickSpec {
...
+                        expect(eventMatcher.loggedEvents).toEventually(haveCount(1), timeout: .seconds(2))
+                    }
+
+                    it("should wait synchronously with registerPreference method") {
+                        let preference = IAMPreferenceBuilder().setUserId("user").build()
+                        queue.async {
+                            iamModule.registerPreference(preference)
+                        }
+                        expect(preferenceRepository.preference).toAfterTimeout(beNil())
+                        expect(preferenceRepository.preference).toEventually(equal(preference), timeout: .seconds(2))
```
malayli, line 187: Is this expect needed?
```
expect(preferenceRepository.preference).toAfterTimeout(beNil())
```

malayli (approved):

Tests/InAppMessagingModuleSpec.swift
imaginaris: Kind of. It checks if preference object wasn't set right after `registerPreference(preference)`. I wanted a slightly different logic to make this more understandable but I couldn't accomplish it with `waitUntil` (I'd had to use DispatchGroup)

PR #80, opened 2021-05-26
fix: fix `flushEventBuffer` logic, fix impressionLeft counter race condition (SDKCF-3821)

imaginaris:
# Description
Fixes a problem when events from `eventBuffer` triggered campaign only once when events allowed to display more times.
Fixes an issue (race condition) when campaigns were validated and added to the queue more times than `maxImpression` allowed

Note: CampaignDispatcher now holds only campaign ID to ensure up-to-date information. The other solution would be to change Campaign model to class type

## Links
SDKCF-3821

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/InAppMessagingModule.swift
+++ b/RInAppMessaging/Classes/InAppMessagingModule.swift
@@ -121,10 +121,10 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
 
     private func flushEventBuffer(discardEvents: Bool) {
         if !discardEvents {
-            eventBuffer.forEach {
-                eventMatcher.matchAndStore(event: $0)
+            eventBuffer.forEach { event in
+                self.eventMatcher.matchAndStore(event: event)
+                self.campaignTriggerAgent.validateAndTriggerCampaigns()
             }
-            campaignTriggerAgent.validateAndTriggerCampaigns()
         }
         eventBuffer.removeAll()
     }
--- a/Tests/PublicAPISpec.swift
+++ b/Tests/PublicAPISpec.swift
@@ -140,12 +140,11 @@ class PublicAPISpec: QuickSpec {
                     configurationManager.simulateRetryDelay = 1.0
                     messageMixerService.delay = 1.0
                     messageMixerService.mockedResponse = TestHelpers.MockResponse.withGeneratedCampaigns(
-                        count: 1, test: false, delay: 100, addContexts: false,
+                        count: 1, test: false, delay: 100, maxImpressions: 2, addContexts: false,
                         triggers: [Trigger(type: .event,
                                            eventType: .loginSuccessful,
                                            eventName: "e1",
                                            attributes: [])])
-                    // maxImpressions = 2
                 })
                 RInAppMessaging.logEvent(LoginSuccessfulEvent())
                 RInAppMessaging.logEvent(LoginSuccessfulEvent())
--- a/Tests/PublicAPISpec.swift
+++ b/Tests/PublicAPISpec.swift
@@ -128,8 +128,13 @@ class PublicAPISpec: QuickSpec {
                                                                pollInterval: .milliseconds(500))
             }
 
-            // it will properly schedule campaigns when events are logged one after another AND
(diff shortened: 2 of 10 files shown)


malayli (approved):

RInAppMessaging/Classes/InAppMessagingModule.swift
```diff
@@ -121,10 +121,10 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
 
     private func flushEventBuffer(discardEvents: Bool) {
         if !discardEvents {
-            eventBuffer.forEach {
-                eventMatcher.matchAndStore(event: $0)
+            eventBuffer.forEach { event in
+                self.eventMatcher.matchAndStore(event: event)
+                self.campaignTriggerAgent.validateAndTriggerCampaigns()
```
donnie-jp, line 126: What's the benefit to bringing this into the loop? Does it fix a problem?

Tests/PublicAPISpec.swift
```diff
@@ -119,15 +118,52 @@ class PublicAPISpec: QuickSpec {
...
+                })
                 RInAppMessaging.logEvent(LoginSuccessfulEvent())
                 expect(eventMatcher.loggedEvents).toAfterTimeout(beEmpty(), timeout: 1)
                 expect(eventMatcher.loggedEvents).toEventually(haveCount(1),
                                                                timeout: .seconds(Int(messageMixerService.delay + 1)),
                                                                pollInterval: .milliseconds(500))
             }
 
+            // it will properly schedule campaigns when events are logged one after another AND
+            it("will display campaigns for matching events that were logged during config retry delay") {
```
donnie-jp: I can't get my head round this test 🤯

donnie-jp (changes requested):
LGTM

Should be targeted to master after #81 is merged

RInAppMessaging/Classes/InAppMessagingModule.swift
imaginaris: Yes. because validator needs to go though the campaign list for each event. those 2 lines are just like a single `logEvent`. I could change the logic inside validator instead. Not sure what's the best

Tests/PublicAPISpec.swift
imaginaris: Should I improve the description or add more context? 🙂

imaginaris:
> LGTM
> 
> Should be targeted to master after #81 is merged

Ok. Rebased and re-targeted to master

RInAppMessaging/Classes/InAppMessagingModule.swift
donnie-jp: No it's fine.

Tests/PublicAPISpec.swift
donnie-jp: Both? 😅

donnie-jp (approved):

Tests/PublicAPISpec.swift
imaginaris: Ok 😄 done. Please check

Tests/PublicAPISpec.swift
```diff
@@ -119,15 +118,58 @@ class PublicAPISpec: QuickSpec {
...
+                    }
+                    return false
+                }), timeout: .seconds(3))
+                expect(UIApplication.shared.keyWindow?.subviews).toEventually(containElementSatisfying({
+                    if let view = $0 as? BaseView {
+                        view.dismiss()
+                        return true
+                    }
+                    return false
+                }))
```
donnie-jp, line 166: So there are 2 expect(contain) because maxImpressions is 2?

Tests/PublicAPISpec.swift
donnie-jp: Much more understandable 👍

Tests/PublicAPISpec.swift
imaginaris: yes, the last expect checks if it doesn't exceed max impression (because there are 3 LoginEvents)

donnie-jp: I understand it now! 🥳

PR #81, opened 2021-05-27
Merge pp support

donnie-jp:
# Description
Merge pp-support branch so that we will only maintain one branch.

## Links
SDKCF-3830

# Checklist


imaginaris (approved):
Nice 👍

imaginaris:
@donnie-jp Can we make SonarCloud skip analysing unit test files? Does it make sense to do that?
(Code duplication percentage is high probably because of the unit tests)

donnie-jp:
> @donnie-jp Can we make SonarCloud skip analysing unit test files? Does it make sense to do that?
> (Code duplication percentage is high probably because of the unit tests)

Yes I think we can (and should) do that. That's my bad I think from when I initially set it up. We can add a sonar config in a separate PR.

PR #82, opened 2021-05-27
fix: fix missing removal of `canOpenURL` (SDKCF-3848)

imaginaris:
# Description
🤦

## Links
SDKCF-3848

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/Views/Presenters/FullViewPresenter.swift
+++ b/RInAppMessaging/Classes/Views/Presenters/FullViewPresenter.swift
@@ -57,9 +57,8 @@ internal class FullViewPresenter: BaseViewPresenter, FullViewPresenterType {
         sendImpressions()
 
         if let unwrappedUri = sender.uri {
-            guard let uriToOpen = URL(string: unwrappedUri),
-                UIApplication.shared.canOpenURL(uriToOpen) else {
 
+            guard let uriToOpen = URL(string: unwrappedUri) else {
                 if let view = view {
                     showURLError(view: view)
                 }
--- a/Tests/ViewPresenterSpec.swift
+++ b/Tests/ViewPresenterSpec.swift
@@ -189,6 +189,59 @@ class ViewPresenterSpec: QuickSpec {
                     presenter.didClickContent()
                     expect(campaignTriggerAgent.wasValidateAndTriggerCampaignsCalled).to(beTrue())
                 }
+
+                context("and content URI is present") {
+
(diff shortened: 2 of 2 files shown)


donnie-jp (approved):
👍

RInAppMessaging/Classes/Views/Presenters/FullViewPresenter.swift
```diff
@@ -57,9 +57,8 @@ internal class FullViewPresenter: BaseViewPresenter, FullViewPresenterType {
         sendImpressions()
 
         if let unwrappedUri = sender.uri {
-            guard let uriToOpen = URL(string: unwrappedUri),
-                UIApplication.shared.canOpenURL(uriToOpen) else {
-
+            
+            guard let uriToOpen = URL(string: unwrappedUri) else {
```
malayli, line 61: Why do we remove the `canOpenURL` check?

imaginaris: Because it fails for https url's on some iOS 14 versions (14.2 for example) where the default browser is Chrome.
And also because if you want to use a deeplink as a url in campaign button, you need to update LSApplicationQueriesSchemes in the app first (If you already released an app you would need to release again).
`openURL` can still fail if there's no app installed that can handle the scheme

RInAppMessaging/Classes/Views/Presenters/FullViewPresenter.swift
```diff
@@ -57,9 +57,8 @@ internal class FullViewPresenter: BaseViewPresenter, FullViewPresenterType {
         sendImpressions()
 
         if let unwrappedUri = sender.uri {
-            guard let uriToOpen = URL(string: unwrappedUri),
-                UIApplication.shared.canOpenURL(uriToOpen) else {
 
+            guard let uriToOpen = URL(string: unwrappedUri) else {
```
donnie-jp, line 61: Why does CodeCov think this isn't covered by tests?

imaginaris: Hmm, it seems Codecov is right. No test reaches this line.
I'll add/modify some tests

imaginaris: I added a few tests for invalid and unsupported url but I didn't add any for positive scenarios because they are hard to test (mocking openURL doesn't make sense and normal openURL breaks other tests)

PR #87, opened 2021-06-21
fix: fix race condition when accessing CampaignRepository (SDKCF-3919)

imaginaris:
# Description
This fixes an issue when impressionLeft data was not updated (and serialized) properly.
The root cause is still unknown. I will investigate that in separate ticket.

## Links
SDKCF-3919

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -74,11 +74,6 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
             return
         }
 
-        CommonUtility.lock(resourcesIn: campaignRepository)
-        defer {
-            CommonUtility.unlock(resourcesIn: campaignRepository)
-        }
-
         let campaignID = queuedCampaignIDs.removeFirst()
         guard let campaign = campaignRepository.list.first(where: { $0.id == campaignID }) else {
             Logger.debug("Warning: Queued campaign \(campaignID) does not exist in the repository anymore")
--- a/RInAppMessaging/Classes/InAppMessagingModule.swift
+++ b/RInAppMessaging/Classes/InAppMessagingModule.swift
@@ -105,10 +105,10 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
 
         let isLogoutOrUserChange = (preferenceRepository.getUserIdentifiers().isEmpty || diff?.isEmpty == false) && !oldUserIdentifiers.isEmpty
         if isLogoutOrUserChange {
-            campaignRepository.resetDataPersistence()
+            campaignRepository.clearLastUserData()
             eventMatcher.clearNonPersistentEvents()
         }
-        campaignRepository.loadCachedData(syncWithLastUserData: false)
+        campaignRepository.loadCachedData(syncWithLastUserData: true)
         campaignsListManager.refreshList()
     }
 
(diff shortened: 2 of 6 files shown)


RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -74,11 +74,6 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
             return
         }
 
-        CommonUtility.lock(resourcesIn: campaignRepository)
```
imaginaris, line 77: This is the actual fix. Why it was causing an issue? - I don't know (yet)

Tests/PublicAPISpec.swift
```diff
@@ -342,6 +338,44 @@ class PublicAPISpec: QuickSpec {
...
+                        $0 is BaseView
+                    })) // wait
+
+                    RInAppMessaging.registerPreference(IAMPreferenceBuilder().setUserId("user").build())
+                    let identifiers = [UserIdentifier(type: .userId, identifier: "user")]
+                    expect(dataCache.getUserData(identifiers: identifiers)?.campaignData?.first?.impressionsLeft)
+                        .toEventually(equal(1))
+                }
+
+                it("will not transfer cached data (sync) from empty user") {
```
imaginaris, line 359: I guess this makes sense(?)

RInAppMessaging/Classes/InAppMessagingModule.swift
```diff
@@ -105,10 +105,10 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
 
         let isLogoutOrUserChange = (preferenceRepository.getUserIdentifiers().isEmpty || diff?.isEmpty == false) && !oldUserIdentifiers.isEmpty
         if isLogoutOrUserChange {
-            campaignRepository.resetDataPersistence()
+            campaignRepository.clearLastUserData()
             eventMatcher.clearNonPersistentEvents()
         }
-        campaignRepository.loadCachedData(syncWithLastUserData: false)
+        campaignRepository.loadCachedData(syncWithLastUserData: true)
```
donnie-jp, line 111: Was this wrong before?
Why change it to true?

Tests/PublicAPISpec.swift
donnie-jp: Can empty user happen in a normal flow?

RInAppMessaging/Classes/CampaignDispatcher.swift
donnie-jp: OK. So this was locked and the `decrementImpressionsLeftInCampaign` failed?

donnie-jp (approved):
LGTM

donnie-jp:
There's a CI test repeatedly failing.

RInAppMessaging/Classes/CampaignDispatcher.swift
imaginaris: I think `decrementImpressionsLeftInCampaign` was called but then the campaign list got overridden by some other part of code (I still have to investigate that, it's hard to debug)

Tests/PublicAPISpec.swift
imaginaris: In DemoApp yes but in general case I think that's not expected behaviour. On the other hand I'm not sure if we should threat empty user as anonymous (nil) user. I think it's a product level decision

RInAppMessaging/Classes/InAppMessagingModule.swift
imaginaris: I think it was kind of wrong before. It could cause problems in some cases. This part covers the 'limbo' state when SDK is initialized and receiving events but user preference is not yet registered. This also fulfils `will transfer cached data (sync) from empty user` test.

donnie-jp (approved):

PR #93, opened 2021-06-30
tests: add a test for calling API methods w/o calling `configure()`

imaginaris:
# Description
Added a test to check if module won't crash if `configure()` is not called

# Checklist

### DIFF
--- a/Tests/PublicAPISpec.swift
+++ b/Tests/PublicAPISpec.swift
@@ -92,11 +92,11 @@ class PublicAPISpec: QuickSpec {
                 let errorDelegate = ErrorDelegate()
                 RInAppMessaging.errorDelegate = errorDelegate
 
-                RInAppMessaging.delegate = delegate
-                RInAppMessaging.accessibilityCompatibleDisplay = true
-                RInAppMessaging.closeMessage(clearQueuedCampaigns: true)
-                RInAppMessaging.logEvent(LoginSuccessfulEvent())
-                RInAppMessaging.registerPreference(IAMPreferenceBuilder().setUserId("user").build())
+                RInAppMessaging.delegate = delegate // 1st error sent
+                RInAppMessaging.accessibilityCompatibleDisplay = true // 2nd error sent
+                RInAppMessaging.closeMessage(clearQueuedCampaigns: true) // 3rd error sent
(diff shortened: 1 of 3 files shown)


Tests/PublicAPISpec.swift
```diff
@@ -72,6 +72,19 @@ class PublicAPISpec: QuickSpec {
 
         describe("RInAppMessaging") {
 
+            it("will not crash if api methods are called prior to configure()") {
+                RInAppMessaging.deinitializeModule()
+                expect(RInAppMessaging.initializedModule).to(beNil())
+
+                let errorDelegate = ErrorDelegate()
+                RInAppMessaging.delegate = delegate
+                RInAppMessaging.errorDelegate = errorDelegate
```
donnie-jp, line 81: Should we call the error delegate if an API is call without the module being initialized?

donnie-jp (approved):
LGTM

Tests/PublicAPISpec.swift
imaginaris: Good idea. Done.

Tests/PublicAPISpec.swift
```diff
@@ -72,6 +72,34 @@ class PublicAPISpec: QuickSpec {
...
+
+                let errorDelegate = ErrorDelegate()
+                RInAppMessaging.errorDelegate = errorDelegate
+
+                RInAppMessaging.delegate = delegate
+                RInAppMessaging.accessibilityCompatibleDisplay = true
+                RInAppMessaging.closeMessage(clearQueuedCampaigns: true)
+                RInAppMessaging.logEvent(LoginSuccessfulEvent())
+                RInAppMessaging.registerPreference(IAMPreferenceBuilder().setUserId("user").build())
+                expect(errorDelegate.totalErrorNumber).toEventually(equal(5))
```
imaginaris, line 100: I'm not sure if it's better to check error description after each call.

donnie-jp: I think this is sufficient but it'd maybe be helpful to have a `// 1st error sent`, `// 2nd` etc. for each API call to make it really obvious where the 5 comes from.

If you want to check the error desc too I'd have a separate test for each API call.

donnie-jp (approved):

PR #95, opened 2021-07-09
SDKCF-3977 avoid Campaign display if image fetch fails

apparition47:
# Description

Match Android behaviour by not displaying the Campaign if its associated `imageUrl` is not nil and fails to download.

Used `URLSession` over `Data(contentsOf:)` since the latter has a unsettable and long default timeout. 

## Links

* SDKCF-3977
* Behaviour reference: [android-inappmessaging DisplayMessageJobIntentService.kt:48-52](https://github.com/rakutentech/android-inappmessaging/blob/73df94dac6f81783c09488a10ff00871432def4b/inappmessaging/src/main/java/com/rakuten/tech/mobile/inappmessaging/runtime/service/DisplayMessageJobIntentService.kt#L49-L52)

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -43,7 +43,7 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
             memoryCapacity: 512_000,
             // response must be <= 5% in order to be cached
             diskCapacity: 400_000_000, // allocation: 5MB * 4 images / 5%
-            diskPath: "RInAppMessaging") 
+            diskPath: "RInAppMessaging")
         httpSession = URLSession(configuration: sessionConfig)
     }
 
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -145,7 +145,6 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
     private func data(from url: URL, completion: @escaping (Data?) -> Void) {
         let sessionConfig = URLSessionConfiguration.default
         sessionConfig.timeoutIntervalForRequest = 5 // seconds
-        sessionConfig.timeoutIntervalForResource = 5
         URLSession(configuration: sessionConfig).dataTask(with: url) { (data, _, error) in
             guard error != nil else {
                 completion(nil)
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -42,7 +42,7 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
         sessionConfig.urlCache = URLCache(
             memoryCapacity: 512_000,
             diskCapacity: 8_000_000_000, // 100_000_000(5MB) * 4 images / 0.05 capacity
-            diskPath: "IAM.CampaignDisplay") // non-iCloud cache path
+            diskPath: "RInAppMessaging") // non-iCloud cache path
         httpSession = URLSession(configuration: sessionConfig)
     }
 
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -27,7 +27,7 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
 
     weak var delegate: CampaignDispatcherDelegate?
     var scheduledTask: DispatchWorkItem?
-    var httpSession: URLSession
+    private(set) var httpSession: URLSession
 
     init(router: RouterType,
          permissionService: DisplayPermissionServiceType,
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -41,8 +41,9 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
         sessionConfig.timeoutIntervalForRequest = Constants.CampaignMessage.imageResourceRequestTimeoutSeconds
         sessionConfig.urlCache = URLCache(
             memoryCapacity: 512_000,
-            diskCapacity: 8_000_000_000, // 100_000_000(5MB) * 4 images / 0.05 capacity
-            diskPath: "RInAppMessaging") // non-iCloud cache path
+            // response must be <= 5% in order to be cached
+            diskCapacity: 8_000_000_000, // allocation: 5MB * 4 images / 5%
+            diskPath: "RInAppMessaging") 
         httpSession = URLSession(configuration: sessionConfig)
     }
 
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -42,7 +42,7 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
         sessionConfig.urlCache = URLCache(
             memoryCapacity: 512_000,
             // response must be <= 5% in order to be cached
-            diskCapacity: 8_000_000_000, // allocation: 5MB * 4 images / 5%
+            diskCapacity: 400_000_000, // allocation: 5MB * 4 images / 5%
             diskPath: "RInAppMessaging") 
         httpSession = URLSession(configuration: sessionConfig)
     }
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -94,14 +94,16 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
             return
         }
 
-        // fetch image from url and display if successful
+        // fetch from imageUrl, display if successful, skip on error
         if let resImgUrlString = campaign.data.messagePayload.resource.imageUrl, let resImgUrl = URL(string: resImgUrlString) {
             data(from: resImgUrl) { imgBlob in
-                guard let imgBlob = imgBlob else {
-                    self.dispatchNext()
-                    return
+                self.dispatchQueue.async {
+                    guard let imgBlob = imgBlob else {
+                        self.dispatchNext()
+                        return
+                    }
+                    self.displayCampaign(campaign, imageBlob: imgBlob)
                 }
-                self.displayCampaign(campaign, imageBlob: imgBlob)
             }
         } else {
             // If no image expected, just display the message.
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -27,6 +27,7 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
 
     weak var delegate: CampaignDispatcherDelegate?
     var scheduledTask: DispatchWorkItem?
+    var httpSession: URLSession
 
     init(router: RouterType,
          permissionService: DisplayPermissionServiceType,
@@ -35,6 +36,10 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
         self.router = router
         self.permissionService = permissionService
         self.campaignRepository = campaignRepository
+
+        let sessionConfig = URLSessionConfiguration.default
+        sessionConfig.timeoutIntervalForRequest = Constants.CampaignMessage.imageResourceRequestTimeoutSeconds
+        httpSession = URLSession(configuration: sessionConfig)
     }
 
     func addToQueue(campaignID: String) {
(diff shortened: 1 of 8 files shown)


RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -89,10 +89,26 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
             return
         }
 
+        // fetch image from url and display if successful
+        if let resImgUrlString = campaign.data.messagePayload.resource.imageUrl, let resImgUrl = URL(string: resImgUrlString) {
```
imaginaris, line 103: I'm wondering if adding additional concurrency here won't cause any unwanted side effects (like race conditions) now or in the future. Should we make image downloading synchronous? Is it worth it?

RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -125,4 +141,16 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
         return campaignData.intervalBetweenDisplaysInMS ??
             Constants.CampaignMessage.defaultIntervalBetweenDisplaysInMS
     }
+
+    private func data(from url: URL, completion: @escaping (Data?) -> Void) {
+        let sessionConfig = URLSessionConfiguration.default
+        sessionConfig.timeoutIntervalForRequest = 5 // seconds
```
imaginaris: I think 5 is reasonable. Let's just put that into a class-level constant (or some `Constants` struct).

RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -125,4 +141,16 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
         return campaignData.intervalBetweenDisplaysInMS ??
             Constants.CampaignMessage.defaultIntervalBetweenDisplaysInMS
     }
+
+    private func data(from url: URL, completion: @escaping (Data?) -> Void) {
+        let sessionConfig = URLSessionConfiguration.default
```
imaginaris: Is default cache policy enough in this case or `returnCacheDataElseLoad` (thinking specifically about images) would better? What do you think?

imaginaris:
I think it'd be nice to have a unit test that checks if campaign is not displayed in case of image download failure. (you can use URLSessionMock for that)

RInAppMessaging/Classes/CampaignDispatcher.swift
apparition47: Forgot that images could be shown more than once. I think relying on the caching policy defined in the HTTP headers as per the default is fairly sensible. But I guess the main problem is that the default `urlCache` size is something low like 512kB(?).

EDIT: Thinking the defaults set on `URLSessionConfiguration.default` are okay 🤔
```
(lldb) po URLSessionConfiguration.default.urlCache!.memoryCapacity
512000

(lldb) po URLSessionConfiguration.default.urlCache!.diskCapacity
10000000
```

RInAppMessaging/Classes/CampaignDispatcher.swift
apparition47: Should be fine if I dispatch the URLSession's completion block the on same queue as where the campaignRepo is being mutated/decremented. Was initially concerned about `CampaignDispatcher` stalling due to `Data(contentsOf:)` long 60s(?) default timeout and slow/intermittent connections

imaginaris: Ok so let's leave as it is. Yeah, the previous implementation wasn't good in terms of error and timeout handling.

RInAppMessaging/Classes/CampaignDispatcher.swift
imaginaris: Yeah it's 512kb for memory cache. 
According to this: https://developer.apple.com/documentation/foundation/nsurlsessiondatadelegate/1411612-urlsession
request can't exceed 5% of the allocated disk capacity to be cached. I think we should increase those values to store maybe 4 campaign images (5MB is the limit for campaign image in the dashboard).

Tests/ReadyCampaignDispatcherSpec.swift
```diff
@@ -21,6 +21,17 @@ class ReadyCampaignDispatcherSpec: QuickSpec {
                                                 permissionService: permissionService,
                                                 campaignRepository: campaignRepository)
                 dispatcher.delegate = delegate
+
+                dispatcher.httpSession = URLSessionMock.mock(originalInstance: .shared)
```
imaginaris: hmm. does that work without calling `URLSessionMock.startMockingURLSession()`?

Tests/ReadyCampaignDispatcherSpec.swift
```diff
@@ -179,6 +190,26 @@ class ReadyCampaignDispatcherSpec: QuickSpec {
...
+                            // swiftlint:disable:next force_cast
+                            let httpSession = dispatcher.httpSession as! URLSessionMock
+                            httpSession.httpResponse = HTTPURLResponse(url: URL(string: "https://example.com/cat.jpg")!,
+                                                                       statusCode: 500,
+                                                                       httpVersion: nil,
+                                                                       headerFields: nil)
+                            httpSession.responseError = NSError(domain: "campaign.error.test", code: 1, userInfo: nil)
+                        }
+
+                        it("won't display campaign when imageUrl is defined") {
```
imaginaris, line 216: Please add also one test for positive case (i.e. image defined and status code 200)

imaginaris (changes requested):

Tests/ReadyCampaignDispatcherSpec.swift
apparition47: Yes, but only because `CampaignDispatcher.httpSession` wasn't private and I was setting it directly from the test. I'll change it to private and just swizzle it out like you did.

RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -39,6 +39,10 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
 
         let sessionConfig = URLSessionConfiguration.default
         sessionConfig.timeoutIntervalForRequest = Constants.CampaignMessage.imageResourceRequestTimeoutSeconds
+        sessionConfig.urlCache = URLCache(
+            memoryCapacity: 512_000,
+            diskCapacity: 8_000_000_000, // 100_000_000(5MB) * 4 images / 0.05 capacity
```
imaginaris: I'm not sure I understand this equation 😄

RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -41,8 +41,9 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
         sessionConfig.timeoutIntervalForRequest = Constants.CampaignMessage.imageResourceRequestTimeoutSeconds
         sessionConfig.urlCache = URLCache(
             memoryCapacity: 512_000,
-            diskCapacity: 8_000_000_000, // 100_000_000(5MB) * 4 images / 0.05 capacity
-            diskPath: "RInAppMessaging") // non-iCloud cache path
+            // response must be <= 5% in order to be cached
+            diskCapacity: 8_000_000_000, // allocation: 5MB * 4 images / 5%
```
imaginaris: 5 000 000*4/0.05 is 400 000 000. Where did 8000M came from?

apparition47: hmmmm 😵

imaginaris (approved):

RInAppMessaging/Classes/Views/Presenters/BaseViewPresenter.swift
```diff
@@ -20,16 +19,7 @@ internal class BaseViewPresenter: BaseViewPresenterType {
...
-        guard let imageURLString = campaign.data.messagePayload.resource.imageUrl,
-            let encodedImageURLString = imageURLString.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed),
-            let encodedImageURL = URL(string: encodedImageURLString),
-            let imageData = try? Data(contentsOf: encodedImageURL) else {
-
-            return nil
-        }
-        return UIImage(data: imageData)
-    }()
+    var associatedImage: UIImage?
```
donnie-jp, line 22: I liked that we fetched this lazily before.

I think now we risk delaying the campaign display to a point where it could be out of context of what the app is now doing.

RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -89,10 +99,28 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
             return
         }
 
+        // fetch from imageUrl, display if successful, skip on error
+        if let resImgUrlString = campaign.data.messagePayload.resource.imageUrl, let resImgUrl = URL(string: resImgUrlString) {
+            data(from: resImgUrl) { imgBlob in
+                self.dispatchQueue.async {
```
donnie-jp, line 105: As this is already executed on the background IAM dispatchQueue why is it necessary to make this async?

RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -35,6 +36,15 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
         self.router = router
         self.permissionService = permissionService
         self.campaignRepository = campaignRepository
+
+        let sessionConfig = URLSessionConfiguration.default
+        sessionConfig.timeoutIntervalForRequest = Constants.CampaignMessage.imageResourceRequestTimeoutSeconds
```
donnie-jp, line 41: 👍 
Definitely better with this lower timeout.

RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -35,6 +36,15 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
...
+        sessionConfig.urlCache = URLCache(
+            memoryCapacity: 512_000,
+            // response must be <= 5% in order to be cached
+            diskCapacity: 400_000_000, // allocation: 5MB * 4 images / 5%
```
donnie-jp, line 45: 400 MB on-disk capacity seems excessive?

https://developer.apple.com/documentation/foundation/urlcache/3329590-init
> A disk cache measured in the tens of megabytes is acceptable in most cases.

RInAppMessaging/Classes/Views/Presenters/BaseViewPresenter.swift
apparition47: This change though is similar to how [Android downloads then caches the image](https://github.com/rakutentech/android-inappmessaging/blob/73df94dac6f81783c09488a10ff00871432def4b/inappmessaging/src/main/java/com/rakuten/tech/mobile/inappmessaging/runtime/service/DisplayMessageJobIntentService.kt#L61) before opening up the view. We could even argue this new way is even **faster (i.e. less delay) than before**, at least for the `maxImpressions>1` case, due to the cache being used here

imaginaris: @donnie-jp previously it was also delayed. `associatedImage` was called before displaying and `UIImage(data: imageData)` is a synchronous method. Now it's just another async call on the same queue

RInAppMessaging/Classes/CampaignDispatcher.swift
imaginaris: True, 400MB is a lot but I don't think it will be used. Maybe instead of increasing disk capacity we could rely on increased memory capacity and bypass that 5% check by setting cache policy to `.returnCacheDataElseLoad` (I'm not sure if that's going to work though)

RInAppMessaging/Classes/Views/Presenters/BaseViewPresenter.swift
donnie-jp: Ah I forgot that the point was to avoid displaying the campaign if the image can't be loaded 😐

RInAppMessaging/Classes/CampaignDispatcher.swift
donnie-jp: Ok, but I'd rather stick to the suggested tens of MB guideline, and set maybe 40MB or 50MB.
Will that not be sufficient?

If you agree, please make this change in a follow-up PR @apparition47

RInAppMessaging/Classes/CampaignDispatcher.swift
apparition47: (Context: this line `data(from: resImgUrl) { imgBlob in` uses URLSession underneath and `URLSessionConfiguration.default`.)

I always thought the URLSession's completion returned on its own background queue so that's why I dispatched it back onto our `CampaignDispatcher.dispatchQueue` where everything else is being run off of.

apparition47: [Completion runs on a separate queue according to SO](https://stackoverflow.com/questions/48951961/how-to-tell-what-queue-a-completionhandler-executes-on). @imaginaris suggested running an experiment and this was the output:
```swift
import Foundation
func currentQueueName() -> String? {
    String(cString: __dispatch_queue_get_label(nil), encoding: .utf8)
}
let dispatchQueue = DispatchQueue(label: "IAM.CampaignDisplay", qos: .userInteractive)
dispatchQueue.async {
    print(currentQueueName()!) // output: IAM.CampaignDisplay
    URLSession.shared.dataTask(with: URL(string:"https://www.google.com")!) { x,y,z in
        print(currentQueueName()!) // output: com.apple.NSURLSession-delegate
    }.resume()
}
```
So I guess we do need to send the completion back to `CampaignDispatcher.dispatchQueue`.

donnie-jp: Yep, you are right

PR #96, opened 2021-07-12
fix: remove error notification for API properties

imaginaris:
# Description
`notifyIfModuleNotInitialized()` is not necessary for API properties because their values are passed to the module during configuration. This change also fixes unwanted notifications when those properties are set just after `configure()`.
Added synchronization for `accessibilityCompatibleDisplay` to avoid crashes.
Removed `.barrier` flag from inAppQueue calls because that flag is applicable only to concurrent queues.

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/RInAppMessaging.swift
+++ b/RInAppMessaging/Classes/RInAppMessaging.swift
@@ -32,8 +32,9 @@ import Foundation
     /// - Note: There is a possibility that changing this property will cause campaigns to display incorrectly
     @objc public static var accessibilityCompatibleDisplay = false {
         didSet {
-            notifyIfModuleNotInitialized()
-            dependencyManager?.resolve(type: RouterType.self)?.accessibilityCompatibleDisplay = accessibilityCompatibleDisplay
+            inAppQueue.async {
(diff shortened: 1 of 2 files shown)


RInAppMessaging/Classes/RInAppMessaging.swift
```diff
@@ -43,10 +44,7 @@ import Foundation
     /// Optional delegate for advanced features
     @objc public static weak var delegate: RInAppMessagingDelegate? {
         didSet {
-            inAppQueue.async {
-                notifyIfModuleNotInitialized()
-                initializedModule?.delegate = delegate
-            }
+            initializedModule?.delegate = delegate
```
donnie-jp, line 47: Why was the queue wrapping removed?
Because you're just setting a property?

donnie-jp (approved):
LGTM

RInAppMessaging/Classes/RInAppMessaging.swift
imaginaris: It wasn't wrapped before. I think it should be safe (I can't remember why I wrapped that in my last PR 😞 )

PR #97, opened 2021-07-12
improve: increase touch area of campaign's close button "x" (SDKCF-3958)

imaginaris:
# Description
Maybe it's not a very clean solution to use `hitArea()` method for handling touches but it was the easiest one.
Changing button's drawing rect would result in adding custom drawing code for button's content, and constraint calculation would operate in negative numbers being totally unintuitive.

## Links
SDKCF-3958

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/Views/FullView.swift
+++ b/RInAppMessaging/Classes/Views/FullView.swift
@@ -22,6 +22,7 @@ internal class FullView: UIView, FullViewType, RichContentBrowsable {
         var exitButtonFontSize: CGFloat = 13 // Size of the exit button.
         var exitButtonSize: CGFloat = 15 // Size of the exit button.
         var exitButtonVerticalOffset: CGFloat = 16 // Position of where the button should be relative to the safe area frame.
+        var exitButtonTouchAreaSize: CGFloat = 44 // Clickable area of exit button used in hitTest
         var dialogViewHorizontalMargin: CGFloat = 20 // The spacing between dialog view and the children elements.
         var dialogViewWidthOffset: CGFloat = 0 // Spacing on the left and right side of subviews.
         var dialogViewWidthMultiplier: CGFloat = 1 // Spacing on the left and right side of subviews.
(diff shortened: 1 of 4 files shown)


RInAppMessaging/Classes/Views/FullView.swift
```diff
@@ -109,6 +110,17 @@ internal class FullView: UIView, FullViewType, RichContentBrowsable {
         bodyViewOffsetYConstraint.constant = hasImage ? 0 : uiConstants.bodyViewSafeAreaOffsetY
     }
 
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        let exitButtonCenter = convert(exitButton.center, from: exitButton.superview)
+        let exitButtonTouchArea = CGRect(origin: exitButtonCenter, size: .zero)
+            .insetBy(dx: -uiConstants.exitButtonTouchAreaSize / 2.0, dy: -uiConstants.exitButtonTouchAreaSize / 2.0)
+
+        if exitButtonTouchArea.contains(point) {
+            return exitButton
```
donnie-jp, line 115: I don't really like that there's near-identical duplicated code in our view classes.
Can it go into a common subclass of UIView?

RInAppMessaging/Classes/Views/FullView.swift
```diff
@@ -109,6 +110,17 @@ internal class FullView: UIView, FullViewType, RichContentBrowsable {
         bodyViewOffsetYConstraint.constant = hasImage ? 0 : uiConstants.bodyViewSafeAreaOffsetY
     }
 
+    override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        let exitButtonCenter = convert(exitButton.center, from: exitButton.superview)
```
donnie-jp: Do we have any way to test this code?
Maybe we need UI tests (doesn't need to be in this PR)

imaginaris: Yeah. I think it's time to write some UI tests

RInAppMessaging/Classes/Views/FullView.swift
imaginaris: Yeah I don't like that either. I probably can move most of that code to some common place

donnie-jp: Do you want to do it in this PR or a follow-up?
Cos I know that this change is kind-of wanted urgently for QA regression.

imaginaris: @donnie-jp This PR. I'll updade in a moment

imaginaris: @donnie-jp Is this ok enough now?
I could move whole method to `BaseView` but then I'd have to bring the concept of exit button as well (outlets, constants etc). The other option would be to have some generic `hitTestViews` collection in BaseView that will be used in hitTest() to ensure min touch size (44pt constant defined in BaseView)

donnie-jp (approved):

RInAppMessaging/Classes/Views/FullView.swift
donnie-jp: Extension seems fine. Thanks for updating.

PR #98, opened 2021-07-13
improve: optimize campaign image cache allotment

apparition47:
# Description

Change image caching policy as [per discussion](https://github.com/rakutentech/ios-inappmessaging/pull/95#discussion_r668648531).

# Discussion

[See empirical test](https://github.com/rakutentech/ios-inappmessaging/pull/98#discussion_r669220444) for our use case

For reference: [JUST EAT](https://tech.justeattakeaway.com/2021/02/01/urlcache-default-size-is-not-enough/) uses 4MB memory & 40MB disk. [PSPDFKit suggested 10MB memory & 1GB disk](https://pspdfkit.com/blog/2020/downloading-large-files-with-urlsession/) (I suppose to be able to score a cache hit on beefy files).

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -39,11 +39,10 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
 
         let sessionConfig = URLSessionConfiguration.default
         sessionConfig.timeoutIntervalForRequest = Constants.CampaignMessage.imageResourceRequestTimeoutSeconds
-        sessionConfig.requestCachePolicy = .returnCacheDataElseLoad
         sessionConfig.urlCache = URLCache(
-            memoryCapacity: 10_000_000,
-            // response must be <= 5% in order to be cached
-            diskCapacity: 40_000_000,
+            // response must be <= 5% of mem/disk cap in order to commited to cache
+            memoryCapacity: 512_000, // default
+            diskCapacity: 101*1024*1024,
             diskPath: "RInAppMessaging")
         httpSession = URLSession(configuration: sessionConfig)
     }
--- a/RInAppMessaging/Classes/CampaignDispatcher.swift
+++ b/RInAppMessaging/Classes/CampaignDispatcher.swift
@@ -41,7 +41,7 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
         sessionConfig.timeoutIntervalForRequest = Constants.CampaignMessage.imageResourceRequestTimeoutSeconds
         sessionConfig.urlCache = URLCache(
             // response must be <= 5% of mem/disk cap in order to commited to cache
-            memoryCapacity: 512_000, // default
+            memoryCapacity: URLCache.shared.memoryCapacity,
             diskCapacity: 101*1024*1024,
             diskPath: "RInAppMessaging")
         httpSession = URLSession(configuration: sessionConfig)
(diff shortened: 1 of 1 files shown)


RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -39,10 +39,11 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
 
         let sessionConfig = URLSessionConfiguration.default
         sessionConfig.timeoutIntervalForRequest = Constants.CampaignMessage.imageResourceRequestTimeoutSeconds
+        sessionConfig.requestCachePolicy = .returnCacheDataElseLoad
```
imaginaris: it seems like this setting doesn't help at all.
I ran some tests on this code and it looks like cache works only if `memoryCapacity` or `diskCapacity` are set to 2000% of desired capacity (5% of cached size). So in order for cache to work for 4 images we have to set wither memory or disk size to 400MiB (and yes, we have to count data as a multiple on 1024). As an alternative we could cache images manually using NSCache and use its data instead of making a request if url-key matches

apparition47: Sad we can't use `URLCache` as-is. The problem with `NSCache` is that it's in-memory only. Maybe we can subclass `URLCache` to always store data (up to 100% capacity).

apparition47: I didn't get the same result. From my own testing, it seems like for 5MB files (max size/worst case for the IAM dashboard), if either `memoryCapacity` or `diskCapacity` is set to `101*1024*1024` (101MB) then the file will be cached. So **101MB** should be the magic number for our use case and seems way more reasonable a setting that we can use without doing any custom caching logic.

Test:

```swift
import Foundation
import PlaygroundSupport

print("check this dir to see if 2 files were cached:")
print(PlaygroundSupport.playgroundSharedDataDirectory.absoluteString + "RTest/fsCachedData/")
// delete dir inbetween runs to clear cache

let conf = URLSessionConfiguration.default
// default cache policy seems OK
//conf.requestCachePolicy = .returnCacheDataElseLoad
//conf.requestCachePolicy = .reloadIgnoringLocalCacheData
conf.urlCache = URLCache(
    memoryCapacity: 512_000 /*default but could be 101MB as well or instead of disk*/,
    diskCapacity: 101*1024*1024,
    diskPath: "RTest")
let session = URLSession(configuration: conf)

// 5MB images to rep worst case in IAM
let img1 = URL(string: "https://upload.wikimedia.org/wikipedia/commons/2/2d/Snake_River_%285mb%29.jpg")!
let img2 = URL(string: "https://sample-videos.com/img/Sample-jpg-image-5mb.jpg")!

session.dataTask(with: img1) { data,_,err in
    print("done -- ", err?.localizedDescription ?? "success")
}.resume()

session.dataTask(with: img2) { data,_,err in
    print("done -- ", err?.localizedDescription ?? "success")
}.resume()

// result: 2 files in /Library/Caches/com.app.blah/RTest/fsCachedData/
```

EDIT: For the  `memoryCapacity=101` case, `diskCapacity` still needs to be larger than default (~19MB) to avoid early eviction. At least 30MB from what I see.

RInAppMessaging/Classes/CampaignDispatcher.swift
```diff
@@ -40,9 +40,9 @@ internal class CampaignDispatcher: CampaignDispatcherType, TaskSchedulable {
         let sessionConfig = URLSessionConfiguration.default
         sessionConfig.timeoutIntervalForRequest = Constants.CampaignMessage.imageResourceRequestTimeoutSeconds
         sessionConfig.urlCache = URLCache(
-            memoryCapacity: 512_000,
-            // response must be <= 5% in order to be cached
-            diskCapacity: 400_000_000, // allocation: 5MB * 4 images / 5%
+            // response must be <= 5% of mem/disk cap in order to committed to cache
+            memoryCapacity: URLCache.shared.memoryCapacity,
+            diskCapacity: Int(5266467/0.05), // fits up to 5MB images
```
imaginaris: 5MiB is 5242880 bytes. (The number you used is good for that sample image we were using in the testing).

imaginaris: The rest is ok 👍

donnie-jp (approved):
LGTM
Interesting that the cache capacity investigation got so complicated 🤓

PR #99, opened 2021-07-14
refactor: IAMPreference builder to UserInfoProvider protocol SDKCF-3970

apparition47:
# Description

Align `registerPreferences(_:)` to match [Android's API](https://github.com/rakutentech/android-inappmessaging/blob/master/inappmessaging/src/main/java/com/rakuten/tech/mobile/inappmessaging/runtime/UserInfoProvider.kt).

## Links

SDKCF-3970

# Checklist


RInAppMessaging/Classes/InAppMessagingModule.swift
```diff
@@ -90,20 +90,25 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
         campaignTriggerAgent.validateAndTriggerCampaigns()
     }
 
-    func registerPreference(_ preference: IAMPreference?) {
+    func registerPreference(_ preference: UserInfoProvider?) {
         guard isEnabled else {
             return
         }
 
+        if BundleInfo.applicationId?.starts(with: "jp.co.rakuten") == true, Bundle.tests == nil {
```
imaginaris, line 99: I think it's better to use `AppDelegate.isTestEnvironment` instead of `Bundle.tests == nil`

RInAppMessaging/Classes/Models/UserInfoProvider.swift
```diff
@@ -0,0 +1,35 @@
...
+ * when needed.
+ */
+@objc public protocol UserInfoProvider {
+
+    /**
+     * Only return RAE token if user is logged in. Else return null.
+     *
+     * @return String of RAE token.
+     */
+    var provideRaeToken: String? { get }
```
imaginaris: I'm not sure but 'provide' sounds like a name for method. Maybe 'provided' or just simply 'userId'. I know it's a one-to-one copy of Android interface but we don't have to be that similar.

RInAppMessaging/Classes/Models/UserInfoProvider.swift
```diff
@@ -0,0 +1,35 @@
...
+     *
+     * @return String of the Rakuten ID.
+     */
+    var provideRakutenId: String? { get }
+}
+
+extension UserInfoProvider {
+    // value used to check UserInfoProvider for equality
+    // swiftlint:disable:next legacy_hashing
+    var hashValue: Int {
```
imaginaris: why the legacy approach. Can't we implement Equatable protocol? Or is it about Obj-C compatibility?

RInAppMessaging/Classes/InAppMessagingModule.swift
```diff
@@ -90,20 +90,25 @@ internal class InAppMessagingModule: AnalyticsBroadcaster,
...
-        let diff = preferenceRepository.preference?.diff(preference)
+        let isDiff = preferenceRepository.canUpdateUserInfo(newUserInfo: preference)
         preferenceRepository.setPreference(preference)
 
         guard isInitialized else {
             return
         }
 
-        let isLogoutOrUserChange = (preferenceRepository.getUserIdentifiers().isEmpty || diff?.isEmpty == false) && !oldUserIdentifiers.isEmpty
+        let isLogoutOrUserChange = isDiff && !oldUserIdentifiers.isEmpty
```
imaginaris: `registerPreference()` method will be called once per app's lifecycle so we can't react to changes here. The user state changes should be analysed when components try to read `preferenceRepository` plus when event is logged so we can reload cache and campaign repository - and that is the main difference.

Tests/Helpers/SharedMocks.swift
```diff
@@ -1,5 +1,38 @@
 @testable import RInAppMessaging
 
+class IAMPreference: UserInfoProvider {
```
imaginaris: I know unit tests will take some time to (re)write and the final PR diff will be huge, but without unit tests we won't know if `UserInfoProvider` approach works.

imaginaris (changes requested):

RInAppMessaging/Classes/Models/UserInfoProvider.swift
apparition47: What do you recommend then? `func` or just rename it?

imaginaris: @donnie-jp on iOS we used to use `accessToken` name for that property. Should we keep using that name to avoid confusing customers or use `raeToken` to keep the name consistent across both platforms?

RInAppMessaging/Classes/Models/UserInfoProvider.swift
apparition47: >Or is it about Obj-C compatibility?

```swift
/// Conforms to NSObject and exposed with objc tag to make it work with Obj-c projects.
@objc public class RInAppMessaging: NSObject {
    @objc public static func registerPreference(_ preference: UserInfoProvider?) {
```

Had trouble implementing `Equatable` because of this. Maybe we handle equality with `isEqual(_:)->Bool` off of the `NSObjectProtocol` if we want to keep `@objc`

RInAppMessaging/Classes/Models/UserInfoProvider.swift
imaginaris: maybe `func` to indicate that this object is not a place to store data :)

RInAppMessaging/Classes/InAppMessagingModule.swift
apparition47: Not sure if I'm following; you're suggesting moving the following into `IAMPreferenceRepository.getUserIdentifiers()`?
```swift
        let isLogoutOrUserChange = isDiff && !oldUserIdentifiers.isEmpty
        if isLogoutOrUserChange {
            campaignRepository.clearLastUserData()
            eventMatcher.clearNonPersistentEvents()
        }
        campaignRepository.loadCachedData(syncWithLastUserData: false)
        campaignsListManager.refreshList()
```

imaginaris: Something like that. Or put it in a separate method so there's a way of getting user identifier without user state verification. (To be honest I'm not sure how what would be the best approach. It's quite a complicated change)

RInAppMessaging/Classes/InAppMessagingModule.swift
apparition47: Where is this mysterious `AppDelegate.isTestEnvironment`?

imaginaris: In the Sample app code... 🤦 Please ignore that comment.

RInAppMessaging/Classes/Models/UserInfoProvider.swift
donnie-jp: I don't really like having references to RAE in our open source SDK (not for security reasons but cos it's too "internal") but naming this is tricky.

Maybe this could be `provideOAuthToken` and then in the future we can add `provide{auth scheme name}Token` e.g., `provideIDToken` - what do you think?

But `provideRaeToken` is going to make perfect sense to internal developers. 🤔

imaginaris: @donnie-jp that's a good idea. I think we shouldn't stick to RAE names too much because soon they'll become obsolete

RInAppMessaging/Classes/Repositories/IAMPreferenceRepository.swift
```diff
@@ -1,11 +1,21 @@
...
 internal class IAMPreferenceRepository {
-    private(set) var preference: IAMPreference?
+    private(set) var preference: UserInfoProvider?
 
-    func setPreference(_ preference: IAMPreference?) {
+    func setPreference(_ preference: UserInfoProvider?) {
         self.preference = preference
     }
 
+    func canUpdateUserInfo(newUserInfo: UserInfoProvider?) -> Bool {
```
malayli: Maybe this signature is more swifty:
`func canUpdate(userInfo newUserInfo: UserInfoProvider?) -> Bool`

RInAppMessaging/Classes/Models/UserInfoProvider.swift
malayli: `provide` sounds weird for me too. Could be `getRaeToken `.

donnie-jp (changes requested):
Merging is on hold until after the upcoming (4.1.0 or 4.0.2) release

apparition47:
@imaginaris My changes are failing these 6 tests. `registerPreference` here has a side effect to check for user info changes and clears lastUserData and nonPersistentEvents if the user logs out or changes to another user. So therefore, it seems like we should now be expecting `true` for these 6?

```swift
// class InAppMessagingModuleSpec: QuickSpec {
it("will not clear last user data when user did not log out or change to another user") {
    let prefA1: UserInfoProvider? = nil
    let prefB1: UserInfoProvider? = aUser
    iamModule.registerPreference(prefA1)
    campaignRepository.resetFlags()
    iamModule.registerPreference(prefB1)
    // Test 1 - ⛔️ fails b/c true
    expect(campaignRepository.wasClearLastUserDataCalled).to(beFalse())
    let prefAAA: UserInfoProvider? = nil
    let prefBBB: UserInfoProvider? = IAMPreference()
    iamModule.registerPreference(prefAAA)
    campaignRepository.resetFlags()
    iamModule.registerPreference(prefBBB)
    // Test 2 - ⛔️ fails b/c true
    expect(campaignRepository.wasClearLastUserDataCalled).to(beFalse())
    let prefAAA1: UserInfoProvider? = IAMPreference()
    let prefBBB1: UserInfoProvider? = nil
    iamModule.registerPreference(prefAAA1)
    campaignRepository.resetFlags()
    iamModule.registerPreference(prefBBB1)
    // Test 3 - ⛔️ fails b/c true
    expect(campaignRepository.wasClearLastUserDataCalled).to(beFalse())
it("will not clear event list when user did not log out or change to another user") {
    let prefA1: UserInfoProvider? = nil
    let prefB1: UserInfoProvider? = aUser
    iamModule.registerPreference(prefA1)
    eventMatcher.wasClearNonPersistentEventsCalled = false // reset
    iamModule.registerPreference(prefB1)
    // Test 4 - ⛔️ fails b/c true
    expect(eventMatcher.wasClearNonPersistentEventsCalled).to(beFalse())
    let prefAAA: UserInfoProvider? = nil
    let prefBBB: UserInfoProvider? = IAMPreference()
    iamModule.registerPreference(prefAAA)
    eventMatcher.wasClearNonPersistentEventsCalled = false // reset
    iamModule.registerPreference(prefBBB)
    // Test 5 - ⛔️ fails b/c true
    expect(eventMatcher.wasClearNonPersistentEventsCalled).to(beFalse())
    let prefAAA1: UserInfoProvider? = IAMPreference()
    let prefBBB1: UserInfoProvider? = nil
    iamModule.registerPreference(prefAAA1)
    eventMatcher.wasClearNonPersistentEventsCalled = false // reset
    iamModule.registerPreference(prefBBB1)
    // Test 6 - ⛔️ fails b/c true
    expect(eventMatcher.wasClearNonPersistentEventsCalled).to(beFalse())
```

RInAppMessaging/Classes/Models/UserInfoProvider.swift
```diff
@@ -0,0 +1,75 @@
...
+                UserIdentifier(type: .rakutenId, identifier: rakutenId)
+            )
+        }
+
+        if let userId = getUserId() {
+            userIdentifiers.append(
+                UserIdentifier(type: .userId, identifier: userId)
+            )
+        }
+
```
donnie-jp, line 48: Should this be calling getIDToken() too?

RInAppMessaging/Classes/Models/UserInfoProvider.swift
```diff
@@ -0,0 +1,75 @@
...
+ * when needed.
+ */
+@objc public protocol UserInfoProvider {
+
+    /**
+     * Only return ID token if user is logged in. Else return null.
+     *
+     * @return String of ID token.
+     */
+    func getIDToken() -> String?
```
donnie-jp: Why is it specifically ID token?
Should it be a generic `getAuthToken()` or just `getToken()`?

RInAppMessaging/Classes/Repositories/AccountRepository.swift
```diff
@@ -0,0 +1,59 @@
+/// Repository to hold `UserInfoProvider`.
+protocol AccountRepositoryType {
+    /// Access token from preference object.
+    /// - Returns: Access token as a string (Optional).
+    var idToken: String? { get }
+    var userInfoProvider: UserInfoProvider? { get }
+    var userInfoHash: UserInfoProvider? { get }
+//    var hasLogoutOrUserChanged: Bool { get }
```
malayli: Can it be removed?

RInAppMessaging/Classes/Repositories/AccountRepository.swift
```diff
@@ -0,0 +1,59 @@
...
+                && !userInfoHash.userIdentifiers.isEmpty
+        case (nil, nil):
+            return false
+        default: // one nil
+            return true
+        }
+    }
+}
+
+class AccountRepository: AccountRepositoryType {
```
malayli: Is there a class that subclasses AccountRepository? Or can it be just final?

RInAppMessaging/Classes/Models/UserInfoProvider.swift
apparition47: This part was transplanted from [`IAMPreferenceRepository`](https://github.com/rakutentech/ios-inappmessaging/blob/master/RInAppMessaging/Classes/Repositories/IAMPreferenceRepository.swift#L21-L29) (also see Android analog [`RuntimeUtil.kt:getUserIdentifiers()`](https://github.com/rakutentech/android-inappmessaging/blob/master/inappmessaging/src/main/java/com/rakuten/tech/mobile/inappmessaging/runtime/utils/RuntimeUtil.kt#L79-L84)) so I think the logic is correct

donnie-jp:
Closing after internal discussion. It will be reopened in future.

PR #101, opened 2021-07-18
tests: add UI tests (SDKCF-2246)

imaginaris:
# Description
Added basic UI tests.
Any ideas for more tests or for any improvements are welcome.

## Links
SDKCF-2246

# Checklist


UITests/SlideUpViewSpec.swift
```diff
@@ -0,0 +1,103 @@
...
+                    launchAppIfNecessary(args: "-campaignType slide-up-redirect")
+                    if !iamView.exists {
+                        app.buttons["login_successful"].tap()
+                        expect(iamView.exists).toEventually(beTrue())
+                    }
+                }
+
+                afterEach {
+                    app.activate() // returns from the browser
+                    usleep(250000) // wait for app switch animation to finish
```
apparition47: Super long delay is a good catch-all in terms of reliability but I personally prefer either disabling animations (or speeding them up at least) just to make UI tests faster and more deterministic

UITests/Campaigns/modal-controls.json
```diff
@@ -0,0 +1,88 @@
...
+        "header": "Test Campaign",
+        "titleColor": "#000000",
+        "headerColor": "#000000",
+        "messageBodyColor": "#000000",
+        "backgroundColor": "#ffffff",
+        "frameColor": "#ffffff",
+        "resource":
+        {
+            "assetsUrl": "",
+            "imageUrl": "file:///System/Library/CoreServices/CoreTypes.bundle/Contents/Resources/ClippingPicture.icns",
```
apparition47: Think this file will be around in future versions of macOS? 🤔 Or what if we execute tests on a real iOS device?

UITests/FullScreenViewSpec.swift
```diff
@@ -0,0 +1,211 @@
...
+                    }
+                }
+
+                it("should not display any buttons") {
+                    expect(iamView.buttons["Button0"].exists).to(beFalse())
+                    expect(iamView.buttons["Button1"].exists).to(beFalse())
+                }
+
+                it("should not display opt-out checkbox") {
+                    expect(iamView.buttons["Do not show me this message again"].exists).to(beFalse())
```
apparition47: Will this work if the device is set to Japanese or are we forcing EN locale for our test env?

UITests/SlideUpViewSpec.swift
```diff
@@ -0,0 +1,103 @@
+import XCTest
```
apparition47: `Quick` already imports `XCTest`

apparition47 (changes requested):

UITests/SlideUpViewSpec.swift
imaginaris: You are right!

UITests/FullScreenViewSpec.swift
imaginaris: Tests always run in english. But actually, I can add an identifier and use it instead

UITests/Campaigns/modal-controls.json
imaginaris: Yeah, might not be. On real device it'll probably fail but are we ever going to run those tests on a real device?
I didn't want to put any http url based image. I can try to embed some image in the test bundle but the path will be different on real device as well. Do you have any better idea?

UITests/SlideUpViewSpec.swift
imaginaris: But that's just 0.25s 😞 It's for the system's app switch animation - I don't have any control over it. I know it's not a pretty solution but I could't find any other way... Without this the test runner tries to tap a button when it still "moving"
```
Failed to synthesize event: Failed to scroll to visible (by AX action) Button, label: 'custom_test', error: Error kAXErrorCannotComplete performing AXAction 2003 on element AX element pid: 98091, elementOrHash.elementID: 140614030137392.10. (Underlying Error: Error kAXErrorCannotComplete performing AXAction 2003 on element AX element pid: 98091, elementOrHash.elementID: 140614030137392.10)
```

imaginaris: I'm removing those tests that involve app transition (opening url). They are too unstable

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
```diff
@@ -15,12 +17,12 @@ struct AtomicGetSet<Value> {
             return queue.sync { value }
         }
         set {
-            queue.sync { value = newValue }
+            queue.sync(flags: .barrier) { value = newValue }
```
imaginaris: changing `AtomicGetSet` to class and using concurrent queue fixes intermittent crashes that were happening (I don't fully understand why though)

fastlane/Fastfile
```diff
@@ -14,10 +14,25 @@ platform :ios do
...
+      slack_only_on_failure: true,
+      skip_slack: ENV['PREVIOUS_BUILD_STATUS'] == 'error', # var defined in Bitrise step
+      output_types: 'json-compilation-database,html,junit',
+      output_files: 'compile_commands.json,report.html,report.junit')
+
+    scan(
+      clean: false,
+      skip_build: true,
       output_directory: './artifacts/unit-tests',
       scheme: ENV['REM_FL_TESTS_SCHEME'] || 'Tests',
```
imaginaris, line 31: unit test must be run as the last step in order to generate code coverage report for Danger (we could probably add some custom paths for the reports instead)

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
apparition47: Do you have a stacktrace of the crash(es)? It looks like you've seen [Donny Wals' suggestions](https://www.donnywals.com/why-your-atomic-property-wrapper-doesnt-work-for-collection-types/) to fix the problem. But it still seems our implementation might not do so well with concurrency and collection value-types especially when collection value types are copied by each thread and copies also given out in the property wrapper's `wrappedValue.get` like he mentions

UITests/Campaigns/modal-controls.json
donnie-jp: Can it use an image embedded as a bundle resource?

UITests/SlideUpViewSpec.swift
apparition47: For the system app switching transitions, does the "Reduce Motion" accessibility setting make a difference?

Podfile
```diff
@@ -7,15 +7,28 @@ target 'RInAppMessaging_Example' do
...
+    end
   end
 end
 
 post_install do |installer|
   system("./configure-secrets.sh InAppMessaging #{secrets.join(" ")}")
+  installer.pods_project.targets.each do |target|
+    target.build_configurations.each do |config|
+      if target.name == 'RInAppMessaging'
+        config.build_settings['ENABLE_TESTABILITY'] = 'YES'
```
apparition47: What's this for? [`ENABLE_TESTABILITY=YES` is already set for the parent `RInAppMessaging` project](https://github.com/rakutentech/ios-inappmessaging/blob/master/RInAppMessaging.xcodeproj/project.pbxproj#L702) then it should be inherited by your new `UITest` target

imaginaris: RInAppMessaging target inherits from `Pods` project unfortunately.

RInAppMessaging/Classes/Helpers/UITestHelper.swift
```diff
@@ -0,0 +1,32 @@
+internal struct UITestHelper {
```
donnie-jp: I'm not keen on this being in the SDK folder's source.

UITests/Campaigns/modal-controls.json
donnie-jp: Bump

fastlane/Fastfile
```diff
@@ -14,10 +14,25 @@ platform :ios do
...
-      device: ENV['REM_FL_TESTS_DEVICE'] || 'REM SDK Unit Tests',
+      device: ENV['REM_FL_TESTS_DEVICE'] || 'iPhone 11',
```
donnie-jp, line 32: 👍

donnie-jp (changes requested):
I really like that we're adding UI tests. Nice work 👍 

However, I'm not so comfortable with us interweaving the "testing environment" checks through our SDK code. I'd like to have a more seamless way of achieving that and with clearer separation between real and test implementation. Could we achieve it through overriding or possibly using protocols (with default implementation in SDK code and test-specific one in UI test target)?

UITests/SlideUpViewSpec.swift
imaginaris: I haven't checked that but I have a feeling that it wouldn't help much

UITests/Campaigns/modal-controls.json
imaginaris: Probably. If putting image name as a url is going to work...
But yeah, it is a better idea

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
imaginaris: @apparition47 That's why I tried to combine his AtomicDict and property wrapper which resulting in changing struct to class and adding concurrent queue. It seems like it helps but to be honest I'm not sure what is what at this point. (I'd have to calmly analyse this case one more time).
Here's the stacktrace (crashed on current line):
```
* thread #20, queue = 'IAM.AtomicProperty489)', stop reason = EXC_BAD_ACCESS (code=1, address=0x3be9ea32eab0)
    frame #0: 0x000000010924f2bd libswiftCore.dylib`_swift_release_dealloc + 13
    frame #1: 0x0000000108c0a0bd RInAppMessaging`partial apply for specialized closure #1 in AtomicGetSet.wrappedValue.setter [inlined] generic not re-abstracted specialization <Swift.Dictionary<Swift.String, RInAppMessaging.UserDataCacheContainer>> of closure #1 () -> () in RInAppMessaging.AtomicGetSet.wrappedValue.setter : τ_0_0 at <compiler-generated>:0 [opt]
    frame #2: 0x0000000108c0a09e RInAppMessaging`partial apply for specialized closure #1 in AtomicGetSet.wrappedValue.setter at <compiler-generated>:0 [opt]
    frame #3: 0x0000000108be5dcc RInAppMessaging`thunk for @callee_guaranteed () -> () at <compiler-generated>:0 [opt]
    frame #4: 0x0000000108be5dee RInAppMessaging`thunk for @escaping @callee_guaranteed () -> () at <compiler-generated>:0 [opt]
    frame #5: 0x00007fff201078df libdispatch.dylib`_dispatch_client_callout + 8
    frame #6: 0x00007fff201155e1 libdispatch.dylib`_dispatch_lane_barrier_sync_invoke_and_complete + 94
  * frame #7: 0x0000000108c08e97 RInAppMessaging`specialized AtomicGetSet.wrappedValue.setter(newValue=<unavailable>, self=<unavailable>) at AtomicWrapper.swift:20:19 [opt]
    frame #8: 0x0000000108c62caf RInAppMessaging`UserDataCache.cacheCampaignData(data=0 values, userIdentifiers=<unavailable>, self=0x0000600000786980) at UserDataCache.swift:55:36 [opt]
    frame #9: 0x0000000109e97d94 Tests`merged closure #1 () -> () in closure #1 () -> () in closure #8 () -> () in closure #2 () -> () in Tests.UserDataCacheSpec.spec() -> () + 212
    frame #10: 0x0000000109e97c7e Tests`thunk for @escaping @callee_guaranteed () -> () at <compiler-generated>:0 [opt]
    frame #11: 0x00007fff2010670d libdispatch.dylib`_dispatch_call_block_and_release + 12
    frame #12: 0x00007fff201078df libdispatch.dylib`_dispatch_client_callout + 8
    frame #13: 0x00007fff2010de15 libdispatch.dylib`_dispatch_lane_serial_drain + 715
    frame #14: 0x00007fff2010e98c libdispatch.dylib`_dispatch_lane_invoke + 400
    frame #15: 0x00007fff20118f81 libdispatch.dylib`_dispatch_workloop_worker_thread + 772
    frame #16: 0x00007fff6034045d libsystem_pthread.dylib`_pthread_wqthread + 314
    frame #17: 0x00007fff6033f42f libsystem_pthread.dylib`start_wqthread + 15
```

RInAppMessaging/Classes/Helpers/UITestHelper.swift
imaginaris: I know but to be honest I think this is the least intrusive way of configuring the SDK for UI Tests.
We could use some external tool to mock the endpoints but that's much more complex solution.

imaginaris:
> I really like that we're adding UI tests. Nice work 👍
> 
> However, I'm not so comfortable with us interweaving the "testing environment" checks through our SDK code. I'd like to have a more seamless way of achieving that and with clearer separation between real and test implementation. Could we achieve it through overriding or possibly using protocols (with default implementation in SDK code and test-specific one in UI test target)?

@donnie-jp But in UITests you can't access module's internal classes. I can only use public API which goes against the idea of performing UI Tests.

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
donnie-jp: @imaginaris I'd like this change considered separately. Please extract the AtomicWrapper related changes to another PR and ensure there is a JIRA tracking it. 

I think this change before/after is a good candidate for the automated stress tests we are planning.

RInAppMessaging/Classes/Helpers/UITestHelper.swift
donnie-jp: I think it's a common pattern on Android to use MockWebServer for running (UI) tests and on iOS there's the possibility of using https://github.com/httpswift/swifter - could be a better approach for us?

donnie-jp: I'd be more comfortable if this struct (or the body/contents of the struct preferably) was namespaced under `Environment` enum as `UITesting`

donnie-jp (changes requested):
I think this will be good to go after these tasks are done:
- moving UITestHelper stuff under the Environment namespace
- extracting the AtomicWrapper changes into a separate PR

RInAppMessaging/Classes/Utilities/AtomicWrapper.swift
malayli: I think that this change should also be done in RAnalytics too as it uses AtomicGetSet.

malayli (approved):

UITests/Campaigns/modal-controls.json
imaginaris: There's no way to get a static url to a resource :( I used Shock url mocking instead

RInAppMessaging/Classes/Helpers/UITestHelper.swift
imaginaris: I removed `UITestHelper` completely. The SDK is now entirely UITest-agnostic

imaginaris:
@donnie-jp @malayli @apparition47 
I made some final updates to this PR. I replaced backend mock injection logic with third party solution called Shock.
Thanks to that, UITests are now non intrusive and the SDK is not aware of any new environment.
Please review latest commit whenever you have some time.

donnie-jp (approved):
Nice!
It's great you found a way to not pollute the SDK classes with test-specific code and the Shock mock server looks very useful. 👍

apparition47 (approved):
Great!

PR #102, opened 2021-07-19
fix: fix scroll view horizontal bouncing (SDKCF-4020)

imaginaris:
# Description
Campaign view should be scrollable only vertically.
It seems like on iPads in landscape mode, for some reason, scroll view's content size is miscalculated.
That calculation is a result of `parentView.setNeedsLayout()` in `displayView` function in `BaseView`.
The difference is just 0.5pt (e.g. 414.0 -> 414.5) I don't know where that 0.5 comes from but I assume that it's a rounding issue.
The line that sets correct content size is wrapped in `DispatchQueue.main.async` in order to be executed after all other calculations.

## Links
SDKCF-4020

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/Views/FullView.swift
+++ b/RInAppMessaging/Classes/Views/FullView.swift
@@ -57,6 +57,7 @@ internal class FullView: UIView, FullViewType, RichContentBrowsable {
     @IBOutlet private weak var optOutView: OptOutMessageView!
     @IBOutlet private weak var optOutAndButtonsSpacer: UIView!
     @IBOutlet private weak var buttonsContainer: UIStackView!
+    @IBOutlet private weak var contentScrollView: UIScrollView!
     @IBOutlet private(set) weak var exitButton: ExitButton! {
         didSet {
             exitButton.invertedColors = hasImage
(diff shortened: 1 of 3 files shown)


malayli (approved):

imaginaris:
@apparition47 @donnie-jp I'm merging this now to unblock QA.
But don't hesitate to leave a comment later.

RInAppMessaging/Classes/Views/FullView.swift
```diff
@@ -108,6 +109,12 @@ internal class FullView: UIView, FullViewType, RichContentBrowsable {
         updateUIConstants()
         exitButtonYPositionConstraint.constant = uiConstants.exitButtonVerticalOffset
         bodyViewOffsetYConstraint.constant = hasImage ? 0 : uiConstants.bodyViewSafeAreaOffsetY
+
+        DispatchQueue.main.async { [self] in
+            // Fixes a problem with content size width being set 0.5pt too much
+            // (landscape iPad), resulting in horizontal scroll bouncing.
+            contentScrollView.contentSize.width = contentScrollView.bounds.width
```
donnie-jp, line 116: Should we guard this inside an iPad landscape mode check?
I guess the guard isn't needed if this change works on all devices.

imaginaris: Yeah, it works on all devices. Actually even if the content size is set to zero, it still displays normally (scrolling is not working though). I'm not sure if that problem was also visible on iPhone X Max in landscape for example

PR #104, opened 2021-07-28
fix: integral image geometry SDKCF-4022

apparition47:
# Description

Campaign image view's `intrinsicContentSize` is recalculated with its aspect ratio which ends up with a fractional `CGFloat`. This causes an issue in some cases where the image doesn't 100% fill the width of its container depending on the fraction with the image being 1px too short on the left and right. Should round CG values to the nearest whole point to avoid frames being drawn on pixel boundaries (ref. [NSHipster, Core​Graphics Geometry Primitives](https://nshipster.com/cggeometry/#integrating-rectangles)) so that it fits its container properly.

ya

## Links

SDKCF-4022

Tested against the 6820 × 4552px image in this ticket on iPhone SE (2nd gen) and iPhone 12 Pro Max.

# Checklist

### DIFF
--- a/RInAppMessaging/Classes/Extensions/CGSize+IAM.swift
+++ b/RInAppMessaging/Classes/Extensions/CGSize+IAM.swift
@@ -1,0 +1,6 @@
+extension CGSize {
+    /// returns a `CGSize` with the smallest integer values for its size to prevent drawing on pixel boundaries
+    var integral: CGSize {
+        CGSize(width: ceil(width), height: ceil(height))
+    }
+}
--- a/RInAppMessaging/Classes/Views/Components/FlexibleHeightImageView.swift
+++ b/RInAppMessaging/Classes/Views/Components/FlexibleHeightImageView.swift
@@ -13,7 +13,7 @@ internal class FlexibleHeightImageView: UIImageView {
         let width = super.intrinsicContentSize.width
         let ratio = image.size.height / image.size.width
 
-        return CGSize(width: width, height: bounds.width * ratio)
+        return CGSize(width: width, height: bounds.width * ratio).integral
     }
 
     override func layoutSubviews() {
--- a/Tests/CoreGraphicsExtensionSpec.swift
+++ b/Tests/CoreGraphicsExtensionSpec.swift
@@ -1,0 +1,39 @@
+import Quick
+import Nimble
+@testable import RInAppMessaging
+
+class CoreGraphicsExtensionsSpec: QuickSpec {
+    override func spec() {
+        describe("CGSize+IAM") {
+            context("when calling integral property") {
+                it("will return a CGSize with whole integers") {
+                    let given = CGSize(width: 0.23, height: 2.34)
+                    let ref = CGRect(origin: .zero, size: given).integral.size
+                    expect(ref).to(equal(given.integral))
+                    expect(ref).to(equal(CGSize(width: 1, height: 3)))
+                }
+
+                it("will return a max CGSize max magnitude") {
+                    let given = CGSize(width: CGFloat.greatestFiniteMagnitude, height: CGFloat.greatestFiniteMagnitude)
+                    let ref = CGRect(origin: .zero, size: given).integral.size
+                    expect(ref).to(equal(given.integral))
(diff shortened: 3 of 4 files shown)


RInAppMessaging/Classes/Extensions/CGSize+IAM.swift
```diff
@@ -0,0 +1,6 @@
+extension CGSize {
+    /// returns a `CGSize` with the smallest integer values for its size to prevent drawing on pixel boundaries
+    var integral: CGSize {
```
imaginaris, line 3: Is that as an adjective? (Checking my English)

imaginaris (approved):
So that was the cause. LGTM 👍🏻

RInAppMessaging/Classes/Extensions/CGSize+IAM.swift
apparition47: Used as a noun but I actually pulled this wording from [`CGRect.integral`](https://developer.apple.com/documentation/coregraphics/cgrect/1456348-integral)

RInAppMessaging/Classes/Views/Components/FlexibleHeightImageView.swift
```diff
@@ -13,7 +13,7 @@ internal class FlexibleHeightImageView: UIImageView {
         let width = super.intrinsicContentSize.width
         let ratio = image.size.height / image.size.width
 
-        return CGSize(width: width, height: bounds.width * ratio)
+        return CGSize(width: width, height: bounds.width * ratio).integral
```
donnie-jp, line 16: Is it testable? It seems like it should be.

RInAppMessaging/Classes/Extensions/CGSize+IAM.swift
```diff
@@ -0,0 +1,6 @@
...
+        CGSize(width: ceil(width), height: ceil(height))
```
malayli, line 4: ceil fixes this issue because width and height can contain decimals?

RInAppMessaging/Classes/Views/Components/FlexibleHeightImageView.swift
malayli: I agree if it can be tested. It reminds me the screen resolution issue that occurred on the Analytics SDK.

Tests/CoreGraphicsExtensionSpec.swift
```diff
@@ -0,0 +1,39 @@
...
+                    let ref = CGRect(origin: .zero, size: given).integral.size
+                    expect(ref).to(equal(given.integral))
+                    expect(ref).to(equal(CGSize(width: 1, height: 1)))
+                }
+
+                it("will return a 1x1 CGSize with least non-zero magnitude") {
+                    let given = CGSize(width: CGFloat.leastNonzeroMagnitude, height: CGFloat.leastNonzeroMagnitude)
+                    let ref = CGRect(origin: .zero, size: given).integral.size
+                    expect(ref).to(equal(given.integral))
+                    expect(ref).to(equal(CGSize(width: 1, height: 1)))
```
malayli, line 34: 👍

malayli (approved):

RInAppMessaging/Classes/Views/Components/FlexibleHeightImageView.swift
apparition47: Added a test but not really sure if I'm testing the right thing. Seems like this kind of thing needs UI tests to verify the fix.

donnie-jp (approved):
LGTM

RInAppMessaging/Classes/Extensions/CGSize+IAM.swift
apparition47: Yes, that seemed like it was the issue

RInAppMessaging/Classes/Views/Components/FlexibleHeightImageView.swift
donnie-jp: thanks - logic tests like you've written is what I had in mind

PR #105, opened 2021-08-06
SDKCF-4022 campaign image scaling

apparition47:
# Description
revert #104, opt for aspect fill scaling instead.

Integral frame alignment seemed to fix most cases but it seems that with the combination of images, device resolutions and orientations, there's always still going to be an odd aspect ratio case that, when used to transform a smaller view frame, will cause the resulting frame to be 1pt off due to rounding. With [`scaleAspectFit`, this extra 1pt will be composited transparently](https://developer.apple.com/documentation/uikit/uiview/contentmode/scaleaspectfit). Should use `scaleAspectFill` instead to **fill** in this 1pt space since we're already mostly in the ballpark with the image aspect ratio we've approximated in `FlexibleHeightImageView.intrinsicContentSize` (i.e. there won't be any perceivable image distortion).

Tested on iPhone 12 Pro Max, iPad Air (4th gen), iPad Pro (12.9") landscape and portrait.

scaling

## Links
SDKCF-4022

# Checklist
- [ ] I have read the [contributing guidelines](../CONTRIBUTING.md).


donnie-jp (approved):
Sounds reasonable.
Probably @malayli or @imaginaris should approve prior to merge.

imaginaris (approved):
👍  UI coding - always full of surprises...

PR #111, opened 2021-09-01
docs: move internal auth SDK examples SDKCF-4169

apparition47:
# Description

Move internal SDK usage references/examples for `registerPreference` to our internal docs.

## Links

mitigates SDKCF-4061

# Checklist
- [ ] I have read the [contributing guidelines](../CONTRIBUTING.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,6 @@ To use the module you must set the following values in your app's `Info.plist`:
 | `InAppMessagingAppSubscriptionID` | your_subscription_key |
 | `InAppMessagingConfigurationURL` | Endpoint for fetching the configuration |
 
-⚠️ Targeted campaigns not showing? When using the staging dashboard and internal RAuthentication/Identity SDK's staging environment, you must specify the corresponding staging values here.
-
 
 ## **Enable and disable the SDK remotely**
 We recommend, as good engineering practice, that you integrate with a remote config service so that you can fetch a feature flag, e.g. `Enable_IAM_SDK`, and use its value to dynamically enable/disable the SDK without making an app release. There are many remote config services on the market, both free and paid.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,6 +37,8 @@ To use the module you must set the following values in your app's `Info.plist`:
 | `InAppMessagingAppSubscriptionID` | your_subscription_key |
 | `InAppMessagingConfigurationURL` | Endpoint for fetching the configuration |
 
+⚠️ Targeted campaigns not showing? When using the staging dashboard and internal RAuthentication/Identity SDK's staging environment, you must specify the corresponding staging values here.
```
imaginaris: The question is - is it better to have the info here (near setup info) or in Troubleshooting section?

donnie-jp: It could be something like
> ⚠️ Targeted campaigns not showing? See our [troubleshooting guide](#troubleshooting-campaigns).

donnie-jp: But actually, we don't expose staging dashboard/endpoints to customers do we? I thought for "staging" they just had to set up a different dashboard app group - is that what's meant here? It isn't clear.

donnie-jp (changes requested):
We don't expose staging dashboard/endpoints to customers do we? I thought for "staging" they just had to set up a different dashboard app group - is that what's meant here? It isn't clear.

README.md
apparition47: Going to revert this line for now. Wasn't aware that our staging was internal

donnie-jp (approved):
👍

README.md
```diff
@@ -111,38 +111,16 @@ RInAppMessaging.logEvent(CustomEvent(withName: "any_event_name_here", withCustom
...
-#### Using the internal Identity SDK
-
-```swift
-let preference = IAMPreferenceBuilder()
-    .setIDTrackingIdentifier("998236")
-    .build()
-
-RInAppMessaging.registerPreference(preference)
-```
+Preferences are not persisted so this function needs to be called on every launch.
```
imaginaris, line 123: 👍

imaginaris (approved):

PR #114, opened 2021-10-05
improve: don't cache test campaigns

imaginaris:
# Description
There's no point in caching test campaigns.

# Checklist


donnie-jp (approved):
LGTM

malayli (approved):

RInAppMessaging/Classes/RInAppMessaging.swift
```diff
@@ -123,7 +123,6 @@ import Foundation
     /// - Parameter provider: object that will always contain up-to-date user information.
     @objc public static func registerPreference(_ provider: UserInfoProvider) {
         inAppQueue.async {
-            notifyIfModuleNotInitialized()
```
vincentdelmundo, line 126: Just want to confirm, why is this removed? Is this because `registerPreference` can be set before `configure`?

vincentdelmundo (approved):
Looks good. Just one to confirm 1 thing, please see inline comment.

RInAppMessaging/Classes/RInAppMessaging.swift
imaginaris: That's right. I think it's better to let clients set `UserInfoProvider` even before init/configure to avoid any issues with API calls order. Only `logEvent` requires `configure` to be called before.

PR #115, opened 2021-10-07
feat: integrate RSDKUtils dependency

imaginaris:
# Description
Integrated RSDKUtils and removed duplicate classes and tests from IAM SDK and Sample app.

This PR should be merged AFTER release of RSDKUtils 2.0.0

## Links
SDKCF-4351

# Checklist


Podfile
```diff
@@ -6,6 +6,9 @@ secrets = ["RIAM_CONFIG_URL", "RIAM_APP_SUBSCRIPTION_KEY"]
 target 'RInAppMessaging_Example' do
   pod 'RInAppMessaging', :path => '.'
   pod 'SwiftLint', '~> 0.42'
+  pod 'RSDKUtils', :git => 'https://github.com/rakutentech/ios-sdkutils.git', :branch => 'release/2.0.0'
+  pod 'RSDKUtils/Nimble', :git => 'https://github.com/rakutentech/ios-sdkutils.git', :branch => 'release/2.0.0'
```
imaginaris: `release/2.0.0` is temporary.
I don't understand why I had to link Nimble and TestHelpers submodules to the main target. When I put them in `Tests-Common` target I got runtime error that Test bundle cannot be loaded because symbols are missing in RSDKUtils framework. do you have any idea why this is happening?

donnie-jp: We need to be certain that we don't put this requirement onto apps using IAM.

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
```diff
@@ -1,3 +1,6 @@
+import Foundation
+import RSDKUtils
```
donnie-jp, line 2: Why is it sometimes `import` like this one and sometimes `import class` ?

donnie-jp (approved):
LGTM

RInAppMessaging/Classes/Repositories/CampaignRepository.swift
imaginaris: When there's only one RSDKUtils component needed for given IAM file the I put `import class/struct/protocol`. If more than one - simple `import RSDKUtils`. I think that way it's easier to 'read the interface of a file'

malayli (approved):

PR #117, opened 2021-10-21
ci: xcresulttool GitHub Actions inline test report (SDKCF-5616)

apparition47:
Add GitHub Check with inline test report using [GitHub Actions xcresulttool](https://github.com/marketplace/actions/xcresulttool))

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,42 @@
+name: Build and Test
+
+on:
+  pull_request_target:
+
+permissions:
+  checks: write
+
+jobs:
+  build-test:
+    runs-on: macOS-12
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.pull_request.head.sha || github.sha }}
+    - run: |
+        bundle install
+        bundle exec pod install
+        xcodebuild \
+        -workspace RInAppMessaging.xcworkspace \
+        -scheme UITests \
+        -resultBundlePath artifacts/ui-tests/UITests \
+        -destination "platform=iOS Simulator,name=iPhone 11" \
+        test
+        xcodebuild \
+        -workspace RInAppMessaging.xcworkspace \
+        -scheme RInAppMessaging-Example \
+        -resultBundlePath artifacts/unit-tests/RInAppMessaging-Example \
+        -destination "platform=iOS Simulator,name=iPhone 11" \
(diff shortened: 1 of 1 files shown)


imaginaris (changes requested):
It looks like Danger can't find .xccoverage files anymore. I think you need to update the Dangerfile

bitrise.yml
```diff
@@ -155,6 +155,21 @@ workflows:
             done
 
             bash codecov -J '^RInAppMessaging$' -X gcov
+    - script@1:
+        title: Merge xcresults for Bitrise Test Reports
+        inputs:
+        - content: |-
+            #!/usr/bin/env bash
+            set -ex
+            xcrun xcresulttool merge artifacts/unit-tests/RInAppMessaging-Example.xcresult artifacts/ui-tests/UITests.xcresult --output-path $BITRISE_DEPLOY_DIR/merged.xcresult
```
imaginaris: Can we use this `merged.xcresult` file in xcov settings in Dangerfile?
I'm wondering why codecov coverage decreased by ~2%?

imaginaris: Ok. Ignore that sentence about Codecov. The coverage is fine - same as on master branch.

imaginaris (approved):

Dangerfile
```diff
@@ -4,6 +4,7 @@ warn("Big PR") if git.lines_of_code > 1000
 xcov.report(
   workspace: 'RInAppMessaging.xcworkspace',
   scheme: 'RInAppMessaging-Example',
+  xccov_file_direct_path: 'artifacts/unit-tests/RInAppMessaging-Example.xcresult',
```
imaginaris: (Idea) Maybe this can be passed as an argument in Danger command (I'm not sure if that's possible). That way we could use the merged .xcresult

apparition47: Maybe possible through env vars? I didn't include the combined xcresult though because the UI test's xcresult isn't set to generate a coverage report

imaginaris: Ah right. Maybe I didn't turn the generation on because it was messing up total coverage. For now the current settings are ok

malayli (approved):

donnie-jp:
@apparition47 can you try to get this working please? 🥺

malayli (approved):

imaginaris (approved):

PR #120, opened 2021-10-29
feat: add SPM support (SDKCF-4388)

imaginaris:
# Description
* Added SPM support
* Added SampleSPM project which uses the same codebase as standard Sample app
* Added new Fastlane lane to build SPM version of RInAppMessaging

## Details
* The SPM module does not contain any Info.plist or anything that contains version number so unfortunately I had to add a resource file which stores the version and must be updated manually.
* None of the test targets can work as Swift Package tests because:
    - they require host app as a dependency
    - use custom .xcconfig files (secrets)
    - UI Tests are not supported
    
SampleSPM works just like normal Sample app. It requires `configure-secrets.sh` script to be run beforehand (either manually or as a part of `pod install`).
I didn't add any new CI step that runs tests in SPM mode because it would significantly increase build time on Bitrise.

## Links
SDKCF-4388

# Checklist


SampleSPM/Example-Debug.xcconfig
```diff
@@ -0,0 +1,2 @@
+// IMPORTANT: Secrets MUST be added to Example-Secrets.xcconfig and that file MUST NOT be added to git.
```
donnie-jp: InAppMessaging-Secrets instead of Example-Secrets?

Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
```diff
@@ -1,3 +1,5 @@
+import UIKit
```
malayli, line 1: `import UIKit.UIColor` ?

Sources/RInAppMessaging/Extensions/UILabel+IAM.swift
```diff
@@ -1,3 +1,5 @@
+import UIKit
```
malayli, line 1: `import UIKit.UILabel` ?

malayli (approved):

Sources/RInAppMessaging/Utilities/BundleInfo.swift
```diff
@@ -0,0 +1,58 @@
...
+    class var inAppConfigurationURL: String? {
+        return Bundle.main.infoDictionary?[Constants.Info.configurationURLKey] as? String
+    }
+}
+
+internal extension Bundle {
+
+    static var sdkAssets: Bundle? {
+        #if SWIFT_PACKAGE
+        module
```
ra-aaron-lee, line 41: Nice

apparition47 (approved):
Nice

Kinda odd when I add this SPM to a project, it adds a bunch of test SPM dependencies too (Nimble, Quick, etc). Maybe related to the SDK Utils?

imaginaris:
> Nice
> 
> Kinda odd when I add this SPM to a project, it adds a bunch of test SPM dependencies too (Nimble, Quick, etc). Maybe related to the SDK Utils?

Yes, they come from SDK Utils. Linked only it its test targets

Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
imaginaris: I would have to import `UIStatusBarStyle` as well (for `.lightContent` part).

Sources/RInAppMessaging/Extensions/UILabel+IAM.swift
imaginaris: Then I'd have to add explicit import for CGFloat and NSMutableParagraphStyle...

donnie-jp (approved):
LGTM

PR #123, opened 2021-11-08
fix: set `triggers` property as optional again (for test campaigns) (SDKCF-4525)

imaginaris:
# Description
The test campaigns payload doesn't have `triggers` property so, at least for now, `triggers` is made optional again.
Unit tests are still using empty value as a default value.

## Links
SDKCF-4525

# Checklist


donnie-jp (approved):
LGTM

PR #130, opened 2021-11-16
SDKCF-4471 Campaign UI design refresh

apparition47:
Update Campaign UI to match current designs.

## Screenshots

See SDKCF-4471.


Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
```diff
@@ -10,4 +10,19 @@ internal extension UIColor {
             return white.withAlphaComponent(0.4)
         }
     }
+
+    /// Returns perceived brightness value based on [Darel Rex Finley's HSP Colour Model](http://alienryderflex.com/hsp.html) from 0 to 1.0 (max brightness)
```
imaginaris, line 14: 👍

Sources/RInAppMessaging/Views/FullView.swift
```diff
@@ -210,7 +207,7 @@ internal class FullView: UIView, FullViewType, RichContentBrowsable {
 
         switch mode {
         case .fullScreen:
-            contentView.topAnchor.constraint(equalTo: topAnchor).isActive = true
```
imaginaris, line 215: Why it was not using `safeAreaLayoutGuide` before? 🤔 I can't remember. Maybe it was for image layout. Anyway let's keep this change

Sources/RInAppMessaging/Views/SlideUpView.swift
```diff
@@ -5,9 +5,9 @@ internal class SlideUpView: UIView, SlideUpViewType {
 
     enum UIConstants {
         static let bodyMessageLabelFontSize: CGFloat = 14
-        static let exitButtonSize: CGFloat = 20
-        static let exitButtonTopMargin: CGFloat = 12
-        static let exitButtonRightMargin: CGFloat = 16
+        static let exitButtonSize: CGFloat = 44
+        static let exitButtonTopMargin: CGFloat = 4
+        static let exitButtonRightMargin: CGFloat = 4
         static let exitButtonTouchAreaSize: CGFloat = 44
```
imaginaris: Do we still need this and `hitTest()` override? (Same applies to FullView class)

imaginaris (commented):
Should the size of exit button be the same in landscape mode?

Sources/RInAppMessaging/Views/FullView.swift
apparition47: Do you mean revert this change?

And before the image used to be positioned into the Safe Area but now we have a spacer with the ⓧ above the image:

apparition47:
> Should the size of exit button be the same in landscape mode?

Android seems to be keeping it the same size so I've kept it the same too

Sources/RInAppMessaging/Views/FullView.swift
imaginaris: That's what I thought. Your change is correct

README.md
```diff
@@ -251,19 +251,33 @@ Campaign impressions (displays) are counted locally for each user. Meaning that
...
+<key>InAppMessagingCustomFontNameRegularWeight</key>
+<string>AdventPro-Regular</string>
+<key>InAppMessagingCustomFontNameMediumWeight</key>
+<string>AdventPro-Medium</string>
+
 <key>UIAppFonts</key>
 <array>
-    <string>M+1r-regular.otf</string>
-    <string>M+1r-medium.otf</string>
+    <string>customfont-regular.otf</string>
```
imaginaris, line 285: Does `customfont` refer to 'AdventPro' in this case? Shouldn't we use `AdventPro` or `customfont` everywhere to avoid confusion?

README.md
```diff
@@ -251,19 +251,33 @@ Campaign impressions (displays) are counted locally for each user. Meaning that
...
+    <string>customfont-medium.otf</string>
 </array>
 ```
 
-Fallsback to system fonts if unavailable or unset.
```
imaginaris: Isn't this sentence still valid?

imaginaris (commented):
UI tests are failing. It might be because of close button changes

README.md
apparition47: Trying to distinguish the font "file name" from the font "PostScript name" somehow. The [UIAppFont](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/iPhoneOSKeys.html#//apple_ref/doc/uid/TP40009252-SW18) takes the file name while the `UIFont(name:size:)` takes in the "PostScript name"

README.md
apparition47: Was thinking that that line was redundant since I mentioned this is a optional feature

README.md
imaginaris: Ahh ok, that makes sense then.

README.md
imaginaris: Hmm yeah, maybe you are right. It is kind of obvious from user perspective that the SDK will use system fonts (or its own fonts) if designated ones aren't available.

Sources/RInAppMessaging/Extensions/UIFont+IAM.swift
```diff
@@ -0,0 +1,20 @@
+import UIKit.UIFont
+
+/// Fonts referenced here must be pre-registered by the host application
+internal extension UIFont {
+    class func iamRegular(ofSize fontSize: CGFloat) -> UIFont {
```
malayli: Do you use `class func` in order to override `iamRegular` function?
If not, can you use `static func`?

Sources/RInAppMessaging/Extensions/UIFont+IAM.swift
```diff
@@ -0,0 +1,20 @@
...
+        guard let customFontName = BundleInfo.customFontNameRegularWeight,
+                let customFont = UIFont(name: customFontName, size: fontSize) else {
+            return UIFont.systemFont(ofSize: fontSize)
+        }
+        return customFont
+    }
+
+    class func iamMedium(ofSize fontSize: CGFloat) -> UIFont {
```
malayli: Do you use `class func` in order to override `iamMedium ` function?
If not, can you use `static func`?

Sources/RInAppMessaging/Utilities/BundleInfo.swift
```diff
@@ -32,6 +32,14 @@ internal class BundleInfo {
     class var inAppConfigurationURL: String? {
         return Bundle.main.infoDictionary?[Constants.Info.configurationURLKey] as? String
     }
+    
+    class var customFontNameRegularWeight: String? {
```
malayli: Do you use `class var` in order to override`customFontNameRegularWeight` variable?
If not, can you use `static var`?

Sources/RInAppMessaging/Utilities/BundleInfo.swift
```diff
@@ -32,6 +32,14 @@ internal class BundleInfo {
...
+        Bundle.main.infoDictionary?[Constants.Info.customFontNameRegularKey] as? String
+    }
+
+    class var customFontNameMediumWeight: String? {
```
malayli: Do you use `class var` in order to override`customFontNameMediumWeight ` variable?
If not, can you use `static var`?

Sources/RInAppMessaging/Views/Components/ExitButton.swift
```diff
@@ -3,23 +3,13 @@ import UIKit
 /// The 'X' button used to close campaign's message.
 internal class ExitButton: UIControl {
 
-    private let xLabel = UILabel()
+    private var exitImageView = UIImageView()
```
malayli: Can `exitImageView` be immutable?

apparition47:
@imaginaris Oops forgot about the UITests, thanks.

>❌ Slide_up_campaign_view__with_trigger_action__should_trigger_another_campaign_after_tapping_the_content() in SlideUpViewSpec failed on iPhone 12 mini
>* Assertions: Assertion Failure at SlideUpViewSpec.swift:102: expected to be false, got `<true>`
>
>❌ Slide_up_campaign_view__with_trigger_action__should_close_campaign_after_tapping_the_content() in SlideUpViewSpec failed on iPhone 12 mini
>* Assertions: Assertion Failure at SlideUpViewSpec.swift:97: expected to be false, got `<true>`
>
>❌ Slide_up_campaign_view__with_redirect_action__should_close_campaign_after_tapping_the_content() in SlideUpViewSpec failed on iPhone 12 mini
>* Assertions: Assertion Failure at SlideUpViewSpec.swift:81: expected to be false, got `<true>`

These ones I'm not sure how these are failing; they seem to fail even on the current `master`.

Sources/RInAppMessaging/Views/Components/ExitButton.swift
apparition47: Updated

imaginaris:
> @imaginaris Oops forgot about the UITests, thanks.
> 
> > ❌ Slide_up_campaign_view__with_trigger_action__should_trigger_another_campaign_after_tapping_the_content() in SlideUpViewSpec failed on iPhone 12 mini
> > 
> > * Assertions: Assertion Failure at SlideUpViewSpec.swift:102: expected to be false, got `<true>`
> > 
> > ❌ Slide_up_campaign_view__with_trigger_action__should_close_campaign_after_tapping_the_content() in SlideUpViewSpec failed on iPhone 12 mini
> > 
> > * Assertions: Assertion Failure at SlideUpViewSpec.swift:97: expected to be false, got `<true>`
> > 
> > ❌ Slide_up_campaign_view__with_redirect_action__should_close_campaign_after_tapping_the_content() in SlideUpViewSpec failed on iPhone 12 mini
> > 
> > * Assertions: Assertion Failure at SlideUpViewSpec.swift:81: expected to be false, got `<true>`
> 
> These ones I'm not sure how these are failing; they seem to fail even on the current `master`.

Hmm They do pass on my machine and Bitrise doesn't complain as well 🤔

apparition47:
@imaginaris Oh weird, it passes on a iPhone 8 (used in the fastlane testing lane) but not on a iPhone 12 mini 🤔

imaginaris:
> @imaginaris Oh weird, it passes on a iPhone 8 (used in the fastlane testing lane) but not on a iPhone 12 mini 🤔

Ok. I see now what's wrong.
This should help: https://github.com/rakutentech/ios-inappmessaging/pull/132

imaginaris (approved):

README.md
```diff
@@ -249,6 +249,35 @@ If user (with registered identifier using `registerPerference()`) opts out from
...
+Get the "PostScript name" of your fonts by:
+
+```bash
+$ fc-scan customfont-medium.otf --format "%{postscriptname}\n"
+
+AdventPro-Medium
+```
+
+In your `Info.plist` configuration, set the PostScript names under `InAppMessagingCustomFontNameText` and `InAppMessagingCustomFontNameButton` along with the file names of both fonts under `UIAppFonts`.
+
```
donnie-jp, line 273: Do they need to be in the plist? Could we just pass them in as optional params to `configure()`?

Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
```diff
@@ -10,4 +10,19 @@ internal extension UIColor {
...
+    var brightness: CGFloat {
+        var (r,g,b) = (CGFloat(0), CGFloat(0), CGFloat(0))
+        getRed(&r, green: &g, blue: &b, alpha: nil) // works with RGB, HSB, extendedGray
+        return sqrt(r*r*0.241 + g*g*0.691 + b*b*0.068)
```
donnie-jp: Should these values be defined as constants?

Sources/RInAppMessaging/Extensions/UIFont+IAM.swift
```diff
@@ -0,0 +1,20 @@
+import UIKit.UIFont
+
+/// Fonts referenced here must be pre-registered by the host application
+internal extension UIFont {
```
donnie-jp, line 4: Is it possible to unit test these by embedding a custom free use font?

README.md
apparition47: We could. But I put it in the plist since the config URL is also in the plist too.

Sources/RInAppMessaging/Extensions/UIFont+IAM.swift
apparition47: Added a glyph-less font file for tests but testing is tricky since fonts are globally registered (i.e. could potentially clobber other tests) and there's no way to unregister fonts at the same time with CoreText (implementation not using UIFontDescriptors so we can fallback onto system fonts).

Sources/RInAppMessaging/Utilities/BundleInfo.swift
```diff
@@ -32,6 +32,14 @@ internal class BundleInfo {
     class var inAppConfigurationURL: String? {
         return Bundle.main.infoDictionary?[Constants.Info.configurationURLKey] as? String
     }
+    
+    static var customFontNameText: String? {
+        Bundle.main.infoDictionary?[Constants.Info.customFontNameTextKey] as? String
```
malayli, line 41: Isn't better to make `var customFontNameText: String?` not static so you can call `Bundle.main. customFontNameText`?
That can allow you to mock `customFontNameText` returned value.

malayli (approved):

PR #134, opened 2021-12-14
improve: improve request error handling (SDKCF-4637)

imaginaris:
# Description
Error handling is now in sync with Android

## Links
SDKCF-4637

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Protocols/HttpRequestable.swift
+++ b/Sources/RInAppMessaging/Protocols/HttpRequestable.swift
@@ -13,7 +13,7 @@ internal enum RequestError: Error {
     case missingMetadata
     case missingParameters
     case taskFailed(Error)
-    case httpError(Int, URLResponse?, Data?)
+    case httpError(UInt, URLResponse?, Data?)
     case bodyEncodingError(Error?)
     case urlBuildingError(Error)
     case urlIsNil
@@ -150,7 +150,7 @@ extension HttpRequestable {
             guard let dataToReturn = data,
                   let serverResponse = response as? HTTPURLResponse else {
 
-                completion(.failure(.httpError(statusCode, response, data)))
+                completion(.failure(.httpError(UInt(statusCode), response, data)))
                 let errorMessage = data != nil ? String(data: data!, encoding: .utf8) : ""
                 Logger.debug("HTTP call failed (\(statusCode))\n\(errorMessage ?? "")")
                 return
--- a/Sources/RInAppMessaging/Protocols/HttpRequestable.swift
+++ b/Sources/RInAppMessaging/Protocols/HttpRequestable.swift
@@ -147,8 +147,7 @@ extension HttpRequestable {
 
             let statusCode = (response as? HTTPURLResponse)?.statusCode ?? 0
 
-            guard 100..<400 ~= statusCode,
-                  let dataToReturn = data,
+            guard let dataToReturn = data,
                   let serverResponse = response as? HTTPURLResponse else {
 
                 completion(.failure(.httpError(statusCode, response, data)))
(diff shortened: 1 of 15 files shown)


Sources/RInAppMessaging/Services/MessageMixerService.swift
```diff
@@ -9,6 +9,8 @@ internal enum MessageMixerServiceError: Error {
     case jsonDecodingError(Error)
     case invalidConfiguration
     case tooManyRequestsError
+    case internalServerError(Int)
```
malayli: Can we use `UInt` instead of `Int`? or the numbers can be negative?

Sources/RInAppMessaging/Utilities/ResponseStateMachine.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+internal struct ResponseStateMachine {
+    internal enum State {
+        case success
+        case error(Error)
+    }
+    private(set) var previousState = State.success
+    private(set) var state = State.success
+    private(set) var consecutiveErrorCount = 0
```
malayli, line 10: Maybe we can declare these variables as private(set) internal?

Sources/RInAppMessaging/Utilities/ResponseStateMachine.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+internal struct ResponseStateMachine {
+    internal enum State {
+        case success
+        case error(Error)
+    }
```
malayli, line 7: Looks like a custom Result :)

Sources/RInAppMessaging/Services/MessageMixerService.swift
imaginaris: Yes, using UInt makes more sense 👍

Sources/RInAppMessaging/Utilities/ResponseStateMachine.swift
imaginaris: That's already how it is now. internal is implicit. I'm putting internal in class/enum/struct declarations to make it clear that it's not public. For properties I don't explicitly write internal. Do you think it's better to use explicit declaration?

Sources/RInAppMessaging/Utilities/ResponseStateMachine.swift
imaginaris: Hmm....true...

imaginaris: @malayli  Do you think I should refactor it somehow to use the `Result` type. I don't have any good idea

Sources/RInAppMessaging/Protocols/HttpRequestable.swift
```diff
@@ -13,7 +13,7 @@ internal enum RequestError: Error {
     case missingMetadata
     case missingParameters
     case taskFailed(Error)
-    case httpError(Int, URLResponse?, Data?)
+    case httpError(UInt, URLResponse?, Data?)
```
malayli, line 16: 👍

Sources/RInAppMessaging/Utilities/ResponseStateMachine.swift
malayli: @imaginaris I don't think so except if you can set a value in case of success.

Sources/RInAppMessaging/Utilities/ResponseStateMachine.swift
malayli: Ok. I thought it was mandatory to use `internal` in the IAM project.

malayli (approved):

Sources/RInAppMessaging/Utilities/ResponseStateMachine.swift
imaginaris: @malayli Actually I noticed I don't need an associated value for `error`

Sources/RInAppMessaging/Utilities/ResponseStateMachine.swift
imaginaris: I see. I don't remember the conversation but I think the requirement was just for class/enum/struct/protocol declarations only.

PR #137, opened 2021-12-23
fix: fix request retry conut issue, improve delay testing (SDKCF-4637)

imaginaris:
# Description
Request retry counting didn't count the initial error so there was always one retry call missing.
I improved tests by adding an option to mock initial delay to significantly reduce time of some test cases.

## Links
SDKCF-4637

# Checklist

### DIFF
--- a/Tests/Tests/CampaignsListManagerSpec.swift
+++ b/Tests/Tests/CampaignsListManagerSpec.swift
@@ -34,6 +34,16 @@ class CampaignsListManagerSpec: QuickSpec {
 
                 context("and service error has occured") {
 
+                    beforeEach {
+                        Constants.Retry.Tests.setInitialDelayMS(1000)
+                        Constants.Retry.Tests.setBackOffUpperBoundSeconds(1)
+                    }
+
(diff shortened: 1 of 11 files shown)


Tests/Tests/CampaignsListManagerSpec.swift
```diff
@@ -34,6 +34,16 @@ class CampaignsListManagerSpec: QuickSpec {
 
                 context("and service error has occured") {
 
+                    beforeEach {
+                        Constants.Retry.Tests.setInitialDelayMS(1000)
+                        Constants.Retry.Tests.setBackOffUpperBoundSeconds(1)
+                    }
+
+                    afterEach {
+                        Constants.Retry.Tests.setInitialDelayMS(10000)
```
apparition47, line 43: Do these need to be set both before **and** after?

imaginaris: In `after` I'm just setting delays back to their original value.

apparition47 (approved):

PR #139, opened 2022-01-02
refactor: fix issues reported by SonarCloud (SDKCF-4730)

imaginaris:
# Description
This PR fixes most of the issues reported by SonarCloud and SonarQube.

## Links
SDKCF-4730

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/InAppMessagingModule.swift
+++ b/Sources/RInAppMessaging/InAppMessagingModule.swift
@@ -160,7 +160,7 @@ extension InAppMessagingModule {
 
 // MARK: - ErrorDelegate methods
 extension InAppMessagingModule {
-    func didReceiveError(sender: ErrorReportable, error: NSError) {
+    func didReceive(error: NSError) {
         aggregatedErrorHandler?(error)
     }
 }
--- a/Sources/RInAppMessaging/Models/Trigger.swift
+++ b/Sources/RInAppMessaging/Models/Trigger.swift
@@ -31,7 +31,7 @@ internal struct TriggerAttribute: Codable, Equatable {
         name = try container.decode(String.self, forKey: CodingKeys.name).lowercased()
         value = try container.decode(String.self, forKey: CodingKeys.value).lowercased()
         type = try container.decode(AttributeType.self, forKey: CodingKeys.type)
-        `operator` = try container.decode(AttributeOperator.self, forKey: CodingKeys.operator)
+        self.operator = try container.decode(AttributeOperator.self, forKey: CodingKeys.operator)
     }
 
     init(name: String, value: String, type: AttributeType, `operator`: AttributeOperator) {
--- a/Sources/RInAppMessaging/Router.swift
+++ b/Sources/RInAppMessaging/Router.swift
@@ -62,7 +62,7 @@ internal class Router: RouterType {
         }
 
         displayQueue.async {
-            guard let viewConstructor = self.viewConstructor(for: campaign, associatedImageData: associatedImageData) else {
+            guard let viewConstructor = self.createViewConstructor(for: campaign, associatedImageData: associatedImageData) else {
                 completion(true)
                 return
             }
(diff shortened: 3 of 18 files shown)


Sources/RInAppMessaging/InAppMessagingModule.swift
```diff
@@ -160,7 +160,7 @@ extension InAppMessagingModule {
 
 // MARK: - ErrorDelegate methods
 extension InAppMessagingModule {
-    func didReceiveError(sender: ErrorReportable, error: NSError) {
+    func didReceive(error: NSError) {
```
donnie-jp, line 163: Do we want to force a major version bump with this change to public API?

Sources/RInAppMessaging/Models/Trigger.swift
```diff
@@ -31,7 +31,7 @@ internal struct TriggerAttribute: Codable, Equatable {
         name = try container.decode(String.self, forKey: CodingKeys.name).lowercased()
         value = try container.decode(String.self, forKey: CodingKeys.value).lowercased()
         type = try container.decode(AttributeType.self, forKey: CodingKeys.type)
-        `operator` = try container.decode(AttributeOperator.self, forKey: CodingKeys.operator)
+        self.operator = try container.decode(AttributeOperator.self, forKey: CodingKeys.operator)
```
donnie-jp, line 34: something looks weird with the formatting of this line - not sure if it's just the github diff highlighting

Sources/RInAppMessaging/Router.swift
```diff
@@ -62,65 +62,71 @@ internal class Router: RouterType {
...
+                guard let rootView = UIApplication.shared.getKeyWindow(),
                       rootView.findIAMViewSubview() == nil,
-                      confirmation() == true else {
+                      confirmation() else {
 
                     completion(true)
                     return
                 }
 
+                let view = viewConstructor()
```
donnie-jp, line 79: viewConstructor returns an optional function - i.e. can be nil - but view is accessed directly below `view.show()` and you didn't need to unwrap it. Why is that?

Tests/Tests/ViewSpec.swift
```diff
@@ -26,19 +26,19 @@ class ViewSpec: QuickSpec {
                 it("will save onDismiss action") {
                     waitUntil { done in
                         let onDismiss: (Bool) -> Void = { _ in done() }
-                        view.show(accessibilityCompatible: false, parentView: parentView, onDismiss: onDismiss)
```
donnie-jp, line 29: Why was accessibilityCompatible removed?

donnie-jp (approved):
A couple of comments to respond to but overall LGTM

Sources/RInAppMessaging/InAppMessagingModule.swift
imaginaris: Ah right. I forgot that `ErrorDelegate` is public. I think it's not worth it. I'll revert this change

imaginaris: Wait, `RInAppMessagingErrorDelegate` is public, not `ErrorDelegate`. So there's no public API change here

Sources/RInAppMessaging/Models/Trigger.swift
imaginaris: Hmm.. I don't see anything weird, except coloring

Sources/RInAppMessaging/Router.swift
imaginaris: viewConstructor is unwrapped in guard a few lines above. And when executed it returns `BaseView` as non-optional.
It's probably confusing because `viewConstructor(Campaign, Data?)` returns a closure that is assigned to `viewConstructor` local variable. I'll change the naming

Tests/Tests/ViewSpec.swift
imaginaris: Because it was never used. It's a leftover from refactoring some time ago. Now `accessibilityCompatible` flag is handled by Router, not by view classes.

PR #142, opened 2022-01-06
fix: fix assertion failure when accessing `Bundle.sdkAssets` in some cases

imaginaris:
# Description
It looks like in some cases (maybe when resource bundle is not yet loaded) `resourceURL` is returning a path to .framework instead of `RInAppMessagingResources.bundle` causing a crash (assertion failure).
This situation happened in one of our client's apps.

EDIT: It turns out that after calling `sdkAssets` for the first time it loads another bundle with `RInAppMessaging` in the identifier causing bundle search logic (inside RSDKUtils) to return .framework at first, and then later always .bundle (because it's first on the list of bundles containing 'RInAppMessaging' in the name) 

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Utilities/BundleInfo.swift
+++ b/Sources/RInAppMessaging/Utilities/BundleInfo.swift
@@ -44,7 +44,12 @@ internal extension Bundle {
         guard let sdkBundleURL = sdk?.resourceURL else {
             return nil
         }
-        return .init(url: sdkBundleURL.appendingPathComponent("RInAppMessagingResources.bundle"))
+        // in some cases `resourceURL` returns a path to .framework file and sometimes to the .bundle file inside
+        if sdkBundleURL.pathExtension == "bundle" {
+            return .init(url: sdkBundleURL)
+        } else {
+            return .init(url: sdkBundleURL.appendingPathComponent("RInAppMessagingResources.bundle"))
+        }
         #endif
     }
 
--- a/Sources/RInAppMessaging/Utilities/BundleInfo.swift
+++ b/Sources/RInAppMessaging/Utilities/BundleInfo.swift
@@ -41,15 +41,14 @@ internal extension Bundle {
         module
         #else
 
+        if let resourceBundle = bundle(bundleIdSubstring: "RInAppMessagingResources") {
+            return resourceBundle
+        }
+
         guard let sdkBundleURL = sdk?.resourceURL else {
             return nil
         }
-        // in some cases `resourceURL` returns a path to .framework file and sometimes to the .bundle file inside
-        if sdkBundleURL.pathExtension == "bundle" {
-            return .init(url: sdkBundleURL)
-        } else {
-            return .init(url: sdkBundleURL.appendingPathComponent("RInAppMessagingResources.bundle"))
-        }
+        return .init(url: sdkBundleURL.appendingPathComponent("RInAppMessagingResources.bundle"))
         #endif
     }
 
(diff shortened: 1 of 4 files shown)


apparition47:
>`resourceURL` is returning a path to .framework instead of `RInAppMessagingResources.bundle`

`bundle.resourceURL` should usually return `**/RInAppMessaging.framework` when using Cocoapods with `use_frameworks`. In that case mentioned, how is that person embedding IAM to get `bundle.resourceURL` to return `.bundle`? Wonder what their proj/Podfile looks like.

imaginaris:
> > `resourceURL` is returning a path to .framework instead of `RInAppMessagingResources.bundle`
> 
> `bundle.resourceURL` should usually return `**/RInAppMessaging.framework` when using Cocoapods with `use_frameworks`. In that case mentioned, how is that person embedding IAM to get `bundle.resourceURL` to return `.bundle`? Wonder what their proj/Podfile looks like.

It looks normal :) It was me that detected this issue in their app after I integrated IAM SDK. I need to take a closer look then and see why there was no issue when I run the code in Sample app.

Sources/RInAppMessaging/Utilities/BundleInfo.swift
```diff
@@ -44,7 +44,12 @@ internal extension Bundle {
         guard let sdkBundleURL = sdk?.resourceURL else {
             return nil
         }
-        return .init(url: sdkBundleURL.appendingPathComponent("RInAppMessagingResources.bundle"))
+        // in some cases `resourceURL` returns a path to .framework file and sometimes to the .bundle file inside
+        if sdkBundleURL.pathExtension == "bundle" {
+            return .init(url: sdkBundleURL)
+        } else {
+            return .init(url: sdkBundleURL.appendingPathComponent("RInAppMessagingResources.bundle"))
+        }
```
malayli: You could use pattern matching here, maybe clearer, not mandatory:
```
switch sdkBundleURL.pathExtension {
    case "bundle":
            return .init(url: sdkBundleURL)
    default:
            return .init(url: sdkBundleURL.appendingPathComponent("RInAppMessagingResources.bundle"))
}
```

malayli (approved):

apparition47 (approved):

donnie-jp (approved):

Sources/RInAppMessaging/Utilities/BundleInfo.swift
imaginaris: @malayli Yeah it looks good for me but I think Sonar will complain about this:
https://rules.sonarsource.com/swift/RSPEC-1301

imaginaris:
> > `resourceURL` is returning a path to .framework instead of `RInAppMessagingResources.bundle`
> 
> `bundle.resourceURL` should usually return `**/RInAppMessaging.framework` when using Cocoapods with `use_frameworks`. In that case mentioned, how is that person embedding IAM to get `bundle.resourceURL` to return `.bundle`? Wonder what their proj/Podfile looks like.

@apparition47 I know what causes this. When `sdkAssets` is called for the first time it creates a new bundle with "RInAppMessaging" string in the identifier. The `sdkBundle(name:)` method tries to create `org.cocoapods.<name>` bundle first, and if it fails it returns FIRST bundle from allBundles that contains given name - org.cocoapods.`RInAppMessaging`Resources.
On simulator SDK bundles seems to have `.iphonesimulator` suffix (e.g. `org.cocoapods.RInAppMessaging.iphonesimulator`)

PR #143, opened 2022-01-06
fix: fix race condition when ping is called during retry process (SDKCF-4637)

imaginaris:
# Description
Calling ping `refreshList` during retry process broke delay and state observation logic.

## Links
SDKCF-4637

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Constants.swift
+++ b/Sources/RInAppMessaging/Constants.swift
@@ -53,13 +53,17 @@ internal enum Constants {
         static let retryCount = 3
 
         enum Default {
-            static fileprivate(set) var initialRetryDelayMS = Int32(10000)
+            fileprivate static let defaultInitialRetryDelayMS = Int32(10000)
+            fileprivate(set) static var initialRetryDelayMS = defaultInitialRetryDelayMS
         }
 
         enum Randomized {
-            static fileprivate(set) var initialRetryDelayMS = Int32(60000)
-            static fileprivate(set) var backOffLowerBoundSeconds = Int32(1)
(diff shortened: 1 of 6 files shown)


malayli (approved):

Sources/RInAppMessaging/Constants.swift
```diff
@@ -71,6 +71,12 @@ internal enum Constants {
             static func setBackOffUpperBoundSeconds(_ bound: Int32) {
                 Randomized.backOffUpperBoundSeconds = bound
             }
+
+            static func setDefaults() {
+                Default.initialRetryDelayMS = Int32(10000)
+                Randomized.initialRetryDelayMS = Int32(60000)
+                Randomized.backOffUpperBoundSeconds = Int32(60)
```
malayli: Maybe these 3 values above can be declared as static immutable:
```
static let defaultiInitialRetryDelayMS = Int32(10000)
static let defaultInitialRetryDelayMS = Int32(60000)
static let defaultBackOffUpperBoundSeconds = Int32(60)
```

malayli (approved):

Sources/RInAppMessaging/Constants.swift
imaginaris: Could be. It sounds more proper but it's just for tests. I'm not sure if I should add more lines of code in this case. I think I'll toss a coin 🙂

malayli: `setDefaults()`is only used in the tests target?

imaginaris: yes. And also Tests are the only place where they are modified. I should add a comment

PR #144, opened 2022-01-13
feat: add tooltip campaigns feature (PoC) (SDKCF-4750)

imaginaris:
# Description
This is a draft PR for Tooltip feature

## Links
SDKCF-4750

# Checklist


Sources/RInAppMessaging/Router.swift
```diff
@@ -124,11 +142,90 @@ internal class Router: RouterType {
...
+
+        DispatchQueue.main.async {
+            self.displayedTooltips[identifier]?.removeFromSuperview()
+            self.displayedTooltips[identifier] = nil
+
+            let position = tooltipData.bodyData.position
+            let tooltipView = TooltipView(
+                model: TooltipViewModel(
+                position: position,
+                image: UIImage(data: imageBlob)!,
```
donnie-jp: What if the data passed in isn't valid image data?
Can we avoid this force unwrap?

Sources/RInAppMessaging/Router.swift
```diff
@@ -124,11 +142,90 @@ internal class Router: RouterType {
...
+                if let uriToOpen = URL(string: tooltipData.bodyData.redirectURL ?? "") {
+                    UIApplication.shared.open(uriToOpen)
+                    onClose()
+                }
+            }
+            tooltipView.onExitButtonTap = onClose
+
+            let superview = self.findParentViewForTooltip(targetView)
+            guard superview != targetView else {
+                Logger.debug("Cannot find suitable view for tooltip targeting: \(targetView.description)")
```
donnie-jp: Maybe this should go to the error delegate?

Sources/RInAppMessaging/Router.swift
```diff
@@ -124,11 +142,90 @@ internal class Router: RouterType {
...
+                self.observers.append(WeakWrapper(value: observer))
+            } else {
+                let observer = targetView.observe(\.frame, options: [.new, .old]) { [weak self] layer, change in
+                    self?.updateFrame(targetView: targetView, tooltipView: tooltipView, superview: superview, position: position)
+                    if self?.isTooltipVisible(tooltipView) == true {
+                        becameVisibleHandler(tooltipView)
+                    }
+                }
+                self.observers.append(WeakWrapper(value: observer))
+            }
```
donnie-jp, line 214: The above code blocks seem to be mostly duplicated

Sources/RInAppMessaging/Router.swift
```diff
@@ -124,11 +142,90 @@ internal class Router: RouterType {
...
+            // TOOLTIP: Keep tooltip away from sceen edges
+    //        let distFromRightEdge = view.window!.bounds.maxX - tooltipView.frame.maxX
+    //        if distFromRightEdge < UIConstants.minDistFromEdge {
+    //            tooltipView.frame.origin.x -= UIConstants.minDistFromEdge - distFromRightEdge
+    //        }
+    //
+    //        let distFromLeftEdge = UIConstants.minDistFromEdge - tooltipView.frame.minX
+    //        if distFromLeftEdge < UIConstants.minDistFromEdge {
+    //            tooltipView.frame.origin.x += distFromLeftEdge
+    //        }
```
donnie-jp: Is the above needed?

Sources/RInAppMessaging/Router.swift
```diff
@@ -137,4 +234,92 @@ internal class Router: RouterType {
             return rootView
         }
     }
+
+    private func findParentViewForTooltip(_ sourceView: UIView) -> UIView {
+        let superview = sourceView.superview
+
+        if superview?.isKind(of: UIScrollView.self) == true {
+            return superview!
```
donnie-jp: Refactor to avoid force unwrap

Sources/RInAppMessaging/Router.swift
```diff
@@ -137,4 +234,92 @@ internal class Router: RouterType {
...
+        } else if superview?.isKind(of: UIWindow.self) == true {
+            if accessibilityCompatibleDisplay,
+               let transitionViewClass = NSClassFromString("UITransitionView"),
+               let transitionView = superview?.subviews.first(where: { !$0.isKind(of: transitionViewClass) }) {
```
donnie-jp: Is this logic safe to rely on continuing to work?

Sources/RInAppMessaging/Views/TooltipView.swift
```diff
@@ -0,0 +1,320 @@
...
+    deinit {
+        exitButton.removeFromSuperview()
+    }
+
+    // MARK: - UIView overrides
+
+    override func draw(_ rect: CGRect) {
+
+        imageBgColor.set()
+        let path: UIBezierPath
```
donnie-jp, line 88: Has tooltip display been tried on varying screen sizes?

Sources/RInAppMessaging/Views/TooltipView.swift
```diff
@@ -0,0 +1,320 @@
...
+    func startAutoFadingIfNeeded(seconds: UInt) {
+        guard autoFadeTimer == nil else {
+            return
+        }
+
+        autoFadeTimer = Timer(fire: Date().addingTimeInterval(TimeInterval(seconds)), interval: 0, repeats: false, block: { [weak self] _ in
+            self?.onExitButtonTap()
+        })
+
+        RunLoop.current.add(autoFadeTimer!, forMode: .common)
```
donnie-jp: Shouldn't force unwrap here - can we catch these with SwiftLint?

donnie-jp: oh - the Timer init can't fail. Maybe it's OK then

donnie-jp (approved):
LGTM although it's a lot of code to go through in one PR. 😅

donnie-jp:
Please target this to a `feat/tooltip` branch.

In a follow-up PR please improve the code coverage of this feature.

Sources/RInAppMessaging/Router.swift
imaginaris: True, it shouldn't be like this. I missed that

Sources/RInAppMessaging/Router.swift
imaginaris: good idea

Sources/RInAppMessaging/Router.swift
imaginaris: Future improvement. For now it can be deleted

Sources/RInAppMessaging/Router.swift
imaginaris: It's been working so far for normal campaigns but it is a liability. There's no guarantee that Apple won't change view hierarchy in the next iOS version. I'll add this to tech debt log - there must be a better way to achieve the same result

Sources/RInAppMessaging/Views/TooltipView.swift
imaginaris: I probably tried just on 2 screen sizes (simulators)

Sources/RInAppMessaging/Views/TooltipView.swift
imaginaris: safe but I refactored it so the force unwrap is avioded

Sources/RInAppMessaging/CampaignTriggerAgent.swift
```diff
@@ -22,12 +25,16 @@ internal struct CampaignTriggerAgent: CampaignTriggerAgentType {
...
                 do {
                     try eventMatcher.removeSetOfMatchedEvents(triggeredEvents, for: campaign)
-                    dispatcher.addToQueue(campaignID: campaign.id)
+                    if campaign.isTooltip {
+                        tooltipDispatcher.setNeedsDisplay(tooltip: campaign)
+                    } else {
+                        campaignDispatcher.addToQueue(campaignID: campaign.id)
+                    }
                 } catch {
                     // Campaign is not ready to be displayed
```
malayli, line 34: Is there a logger for errors in IAM? Is is relevant to log the error here?

Sources/RInAppMessaging/Router.swift
```diff
@@ -137,4 +230,96 @@ internal class Router: RouterType {
             return rootView
         }
     }
+
+    private func findParentViewForTooltip(_ sourceView: UIView) -> UIView {
+        guard let superview = sourceView.superview else {
+            return sourceView
+        }
+
```
malayli, line 238: I think you can use pattern matching instead of if/else, it would improve the readability:

Sources/RInAppMessaging/ViewListener.swift
```diff
@@ -0,0 +1,225 @@
...
+            }
+        }
+    }
+
+    private func performSwizzling() -> Bool {
+        [swizzle(#selector(UIView.didMoveToSuperview), with: #selector(UIView.swizzled_didMoveToSuperview)),
+         swizzle(#selector(UIView.removeFromSuperview), with: #selector(UIView.swizzled_removeFromSuperview)),
+         swizzle(#selector(setter: UIView.accessibilityIdentifier),
+                 with: #selector(NSObject.swizzled_setAccessibilityIdentifier),
+                 of: NSObject.self),
```
malayli, line 95: Can the comments below be removed?

Sources/RInAppMessaging/ViewListener.swift
```diff
@@ -0,0 +1,225 @@
...
+                  return false
+              }
+
+        method_exchangeImplementations(originalMethod, swizzledMethod)
+
+        return true
+    }
+}
+
+private extension NSObject {
```
malayli, line 112: Is this swizzling used for tests purpose?

malayli (approved):

Sources/RInAppMessaging/Router.swift
```diff
@@ -176,7 +182,7 @@ internal class Router: RouterType, ViewListenerObserver {
 
             let superview = self.findParentViewForTooltip(targetView)
             guard superview != targetView else {
-                Logger.debug("Cannot find suitable view for tooltip targeting: \(targetView.description)")
+                self.reportError(description: "Invalid image data for tooltip targeting \(tooltipData.bodyData.uiElementIdentifier)", data: targetView)
```
donnie-jp: I think this is the wrong error message

Sources/RInAppMessaging/TooltipDispatcher.swift
```diff
@@ -0,0 +1,134 @@
...
+        self.router = router
+        self.campaignRepository = campaignRepository
+        self.viewListener = viewListener
+
+        let sessionConfig = URLSessionConfiguration.default
+        sessionConfig.timeoutIntervalForRequest = Constants.CampaignMessage.imageRequestTimeoutSeconds
+        sessionConfig.timeoutIntervalForResource = Constants.CampaignMessage.imageResourceTimeoutSeconds
+        sessionConfig.waitsForConnectivity = true
+        sessionConfig.urlCache = URLCache(
+            // response must be <= 5% of mem/disk cap in order to committed to cache
```
apparition47, line 31: Nice 😎

Sources/RInAppMessaging/TooltipDispatcher.swift
```diff
@@ -0,0 +1,134 @@
...
+                identifier: identifier,
+                imageBlob: imageBlob,
+                becameVisibleHandler: { tooltipView in
+                    guard let autoFadeSeconds = tooltipData.bodyData.autoFadeSeconds, autoFadeSeconds > 0 else {
+                        return
+                    }
+                    tooltipView.startAutoFadingIfNeeded(seconds: autoFadeSeconds)
+                }, completion: {
+                    self.campaignRepository.decrementImpressionsLeftInCampaign(id: tooltip.id)
+                    self.queuedTooltips.remove(tooltip)
```
apparition47, line 87: Might be strong reference cycles here?

Sources/RInAppMessaging/CampaignTriggerAgent.swift
imaginaris: There is but this is a valid state - not an error. When `removeSetOfMatchedEvents` fails it means that the campaign still needs more events to be logged

Sources/RInAppMessaging/Router.swift
imaginaris: True. Now it looks much better 👍

Sources/RInAppMessaging/ViewListener.swift
imaginaris: Yes, now they can

Sources/RInAppMessaging/ViewListener.swift
imaginaris: No. it is used for swizzling accessibilityIdentifier which is defined in `UIAccessibilityIdentification` protocol. NSObject is a common class for classes that implement that protocol.

Sources/RInAppMessaging/Router.swift
imaginaris: How did this happen?! 😅

Sources/RInAppMessaging/TooltipDispatcher.swift
imaginaris: router does not store this completion block anywhere so I think it should be ok.

PR #152, opened 2022-01-28
fix: make the swift build target configurable (SDKCF-4790)

donnie-jp:
# Description
- To get the swift build command to succeed on Xcode 12.x we need to set a different target
- This change makes the target configurable by an injectable env var

## Links
SDKCF-4790

# Checklist


donnie-jp:
Closed.
Doesn't seem to be needed.

PR #165, opened 2022-03-27
improve: draw button's border if it has the same bg color as campaign (SDKCF-4859)

imaginaris:
# Description
Action button should have a border when campaign body has the same background color.

## Links
SDKCF-4859

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
+++ b/Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
@@ -16,7 +16,7 @@ internal extension UIColor {
         }
     }
 
-    static let buttonBorderDefaultColor = UIColor(hexString: "#D1D1D1")!
+    static let buttonBorderDefaultColor = UIColor(hexString: "#D1D1D1") ?? .lightGray
 
     /// Returns perceived brightness value based on [Darel Rex Finley's HSP Colour Model](http://alienryderflex.com/hsp.html) from 0 to 1.0 (max brightness)
     var brightness: CGFloat {
--- a/Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
+++ b/Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
@@ -33,4 +33,23 @@ internal extension UIColor {
         return brightness > threshold
     }
 
+    /// Returns distance between colors based on the low-cost approximation algorithm from [](https://www.compuphase.com/cmetric.htm)
+    func distance(from: UIColor) -> Int {
+        var (r1, g1, b1) = (CGFloat(0), CGFloat(0), CGFloat(0))
+        var (r2, g2, b2) = (CGFloat(0), CGFloat(0), CGFloat(0))
+        getRed(&r1, green: &g1, blue: &b1, alpha: nil) // works with RGB, HSB, extendedGray
+        from.getRed(&r2, green: &g2, blue: &b2, alpha: nil)
+
+        let rMean = (r1 * 255 + r2 * 255) / 2
+        let dR = r1 * 255 - r2 * 255
+        let dG = g1 * 255 - g2 * 255
+        let dB = b1 * 255 - b2 * 255
+
+        return Int(round(sqrt((2 + rMean / 256) * dR * dR + 4 * dG * dG + (2 + (255 - rMean) / 256) * dB * dB)))
+    }
+
+    func isComparable(to anotherColor: UIColor) -> Bool {
+        let threshold = 15
+        return distance(from: anotherColor) <= threshold
+    }
 }
--- a/Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
+++ b/Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
@@ -1,4 +1,9 @@
 import UIKit
+#if canImport(RSDKUtilsMain)
+import RSDKUtilsMain // SPM version
+#else
+import RSDKUtils
+#endif
 
 /// Extension to `UIColor` class to provide addtional initializers required by InAppMessaging.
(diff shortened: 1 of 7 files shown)


Sources/RInAppMessaging/Views/Presenters/FullViewPresenter.swift
```diff
@@ -38,14 +41,15 @@ internal class FullViewPresenter: BaseViewPresenter, FullViewPresenterType {
...
             buttonsToAdd.append((
                 ActionButton(impression: index == 0 ? ImpressionType.actionOne : ImpressionType.actionTwo,
                              uri: button.buttonBehavior.uri,
                              trigger: button.campaignTrigger),
                 ActionButtonViewModel(text: button.buttonText,
                                       textColor: UIColor(hexString: button.buttonTextColor) ?? .black,
-                                      backgroundColor: UIColor(hexString: button.buttonBackgroundColor) ?? .white)))
-
+                                      backgroundColor: UIColor(hexString: button.buttonBackgroundColor) ?? .white,
+                                      shouldDrawBorder: viewBackgroundColor == backgroundColor)))
```
apparition47: This means the colours have be **exactly** the same? What if we have a white on off-white? Seems like there should be a degree of tolerance here

Sources/RInAppMessaging/Views/Components/ActionButton.swift
```diff
@@ -8,7 +13,8 @@ internal class ActionButton: UIButton {
         static let minFontSize: CGFloat = 10
         static let labelMargin = UIEdgeInsets(top: 8, left: 8, bottom: 8, right: 8)
         static let cornerRadius: CGFloat = 8
-        static let borderWidth: CGFloat = 0
+        static let borderWidth: CGFloat = 0.5
+        static let borderColor = UIColor(hexString: "#D1D1D1")!
```
apparition47: Is this colour okay if the background colours are dark? What if the modal bg is also grey?

apparition47 (changes requested):

Sources/RInAppMessaging/Views/Presenters/FullViewPresenter.swift
imaginaris: Probably. there are no requirements to have this kind of tolerance. We can't always protect users from their poor UI choices. But I see your point. Let me confirm with PDM

Sources/RInAppMessaging/Views/Components/ActionButton.swift
imaginaris: Your are totally right. I was just following design pattern for one customer and didn't think about other colors. I'll use textColor if bg is not white

Sources/RInAppMessaging/Views/Presenters/FullViewPresenter.swift
imaginaris: @apparition47 PDM confirmed that there should be some tolerance. You win 😉

apparition47: Oh, shouldn't have asked him then 😅

Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
```diff
@@ -26,4 +33,23 @@ internal extension UIColor {
...
+        var (r2, g2, b2) = (CGFloat(0), CGFloat(0), CGFloat(0))
+        getRed(&r1, green: &g1, blue: &b1, alpha: nil) // works with RGB, HSB, extendedGray
+        from.getRed(&r2, green: &g2, blue: &b2, alpha: nil)
+
+        let rMean = (r1 * 255 + r2 * 255) / 2
+        let dR = r1 * 255 - r2 * 255
+        let dG = g1 * 255 - g2 * 255
+        let dB = b1 * 255 - b2 * 255
+
+        return Int(round(sqrt((2 + rMean / 256) * dR * dR + 4 * dG * dG + (2 + (255 - rMean) / 256) * dB * dB)))
```
apparition47, line 48:
```c
return sqrt((((512+rmean)*r*r)>>8) + 4*g*g + (((767-rmean)*b*b)>>8));
```

The C interpretation in the paper used a (supposedly faster) bitshift right op by 8 instead of division in the original formula. Probably an academic thought in this case but wonder if there would've been a big perf difference

apparition47 (approved):
Cool!

Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
```diff
@@ -11,6 +16,8 @@ internal extension UIColor {
         }
     }
 
+    static let buttonBorderDefaultColor = UIColor(hexString: "#D1D1D1")!
```
malayli: Isn't risky to force unwrap `UIColor(hexString: "#D1D1D1")!`?

Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
imaginaris: Yeah, I decided to go with readability because this formula is not going to be often executed. So I think on modern devices and having Swift optimizer, there shouldn't be a big pref difference. I don't have any source to back me up on this though. If that was a C code I would use the bit shift approach 🙂

Sources/RInAppMessaging/Extensions/UIColor+IAM.swift
imaginaris: I don't think so. It's a constant. If it doesn't crash now, it shouldn't crash later. Having `buttonBorderDefaultColor` as optional would be misleading in my opinion. But I could add a nil coalescence with `UIColor.grey` for example just to avoid that force unwrap.

PR #167, opened 2022-04-01
fix: retain impressionsLeft value for test campaigns (SDKCF-5027)

imaginaris:
# Description
In the previous SDKCF-5027 PR I forgot to update sync method in CampaignRepository to retain cached `impressionsLeft` value in test campaigns.

EDIT: Added logic to ignore maxImpressions and infiniteImpressions values

## Links
SDKCF-5027

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Repositories/CampaignRepository.swift
+++ b/Sources/RInAppMessaging/Repositories/CampaignRepository.swift
@@ -69,18 +69,9 @@ internal class CampaignRepository: CampaignRepositoryType {
     func syncWith(list: [Campaign], timestampMilliseconds: Int64) {
         lastSyncInMilliseconds = timestampMilliseconds
         let oldList = campaigns.get()
-        var updatedList = [Campaign]()
-
-        let (testCampaigns, newCampaigns) = list.reduce(into: ([Campaign](), [Campaign]())) { partialResult, campaign in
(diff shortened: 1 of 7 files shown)


Sources/RInAppMessaging/Repositories/CampaignRepository.swift
```diff
@@ -94,10 +85,10 @@ internal class CampaignRepository: CampaignRepositoryType {
                     updatedCampaign = Campaign.updatedCampaign(updatedCampaign, withImpressionLeft: newImpressionsLeft)
                 }
             }
-            updatedList.append(updatedCampaign)
+            return updatedCampaign
         }
-        campaigns.set(value: updatedList + testCampaigns)
-        saveDataToCache(updatedList + testCampaigns)
```
malayli, line 100: The fix was about removing `testCampaigns`?

malayli (approved):

Sources/RInAppMessaging/Repositories/CampaignRepository.swift
imaginaris: The fix was actually about including test campaigns in the update closure above. So now, since test campaigns are basically treated as normal campaigns, there is no need to filter them out to a separate `testCampaigns` var

PR #170, opened 2022-04-11
improve: set default values for new properties in cached campaigns data (SDKCF-5003)

imaginaris:
# Description
Cached campaign model might not contain the new properties (flags) added as a part of this ticket.
In order to avoid decoding exception and clearing cached data, a default value will be set for each property.

## Links
SDKCF-5003
SDKCF-5142

# Checklist


Sources/RInAppMessaging/Models/Responses/Campaign.swift
```diff
@@ -0,0 +1,70 @@
+import struct Foundation.Date
```
imaginaris, line 1: This code has been moved w/o any change from PingResponse.swift

Sources/RInAppMessaging/Models/Responses/CampaignDataModels.swift
```diff
@@ -0,0 +1,71 @@
+internal struct MessagePayload: Codable {
```
imaginaris, line 1: This code has been moved w/o any change from PingResponse.swift

donnie-jp (approved):
LGTM

PR #192, opened 2022-06-20
improve: handle missing config url in release mode to avoid a crash, add config url fallback value for sample app archive

imaginaris:
# Description
In Debug mode missing config URL throws an assertion. In release mode, ConfigurationService cannot be initialized which causes a crash because ConfigurationManager init in `MainContainerFactory` force unwraps the reference.
This PR fixes this issue by properly handling nil values and disabling the SDK in the end.

Added a dummy config url to set in sample app archives in case RIAM_CONFIG_URL is not set (PR builds)
(This PR fixes the issue with Emerge Performance Analysis tool)

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/DependencyManager/MainContainer.swift
+++ b/Sources/RInAppMessaging/DependencyManager/MainContainer.swift
@@ -36,7 +36,7 @@ internal enum MainContainerFactory {
             }),
             ContainerElement(type: ConfigurationManagerType.self, factory: {
                 ConfigurationManager(reachability: manager.resolve(type: ReachabilityType.self),
-                                     configurationService: manager.resolve(type: ConfigurationServiceType.self)!,
+                                     configurationService: manager.resolve(type: ConfigurationServiceType.self),
                                      configurationRepository: manager.resolve(type: ConfigurationRepositoryType.self)!,
                                      resumeQueue: RInAppMessaging.inAppQueue)
             }),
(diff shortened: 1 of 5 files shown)


malayli (approved):

donnie-jp (approved):
LGTM

Sources/RInAppMessaging/DependencyManager/MainContainer.swift
```diff
@@ -36,7 +36,7 @@ internal enum MainContainerFactory {
             }),
             ContainerElement(type: ConfigurationManagerType.self, factory: {
                 ConfigurationManager(reachability: manager.resolve(type: ReachabilityType.self),
-                                     configurationService: manager.resolve(type: ConfigurationServiceType.self)!,
```
donnie-jp, line 39: 🎉 

I've never been a fan of force unwrap as you know 😄

Sources/RInAppMessaging/DependencyManager/MainContainer.swift
```diff
@@ -36,7 +36,7 @@ internal enum MainContainerFactory {
...
+                                     configurationService: manager.resolve(type: ConfigurationServiceType.self),
                                      configurationRepository: manager.resolve(type: ConfigurationRepositoryType.self)!,
```
donnie-jp, line 40: nit: is this force unwrap guaranteed to be safe?

imaginaris: Yes. The init cannot fail

PR #209, opened 2022-07-22
build: update Podfile (Xcode 13.3 support)

imaginaris:
# Description
Changes necessary for any future PR targeted to `feat/tooltip` branch to pass CI build
(The changes are cherry-picked from the master branch)

# Checklist
- [X] I have read the [contributing guidelines](../CONTRIBUTING.md).
- [ ] I wrote/updated tests for new/changed code
- [X] I removed all sensitive data **before every commit**, including API endpoints and keys
- [X] I ran `fastlane ci` without errors

PR #210, opened 2022-07-22
improve: accept any form of `[Tooltip]` title prefix (SDKCF-5638)

imaginaris:
# Description
The SDK is currently set to expect '[ToolTip]' prefix in campaign title using case sensitive comparison. Customers often put '[Tooltip]' in the title instead of expected '[ToolTip]' which causes tooltip campaign to display incorrectly (i.e. as a standard campaign).
'Tooltip' is a proper dictionary form so it should be accepted as well as other forms with different letter case

## Links
SDKCF-5638
PR to be merged first - https://github.com/rakutentech/ios-inappmessaging/pull/209

# Checklist

### DIFF
--- a/Sample/ping.json
+++ b/Sample/ping.json
@@ -17,7 +17,7 @@
                     }
                 ],
                 "messagePayload": {
-                    "title": "[ToolTip] tooltip 1",
+                    "title": "[Tooltip] tooltip 1",
                     "messageBody": "{\"UIElement\" : \"tooltip.test.1\", \"position\": \"bottom-right\"}",
                     "header": "Test Campaign",
                     "titleColor": "#000000",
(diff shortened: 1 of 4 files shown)


apparition47 (approved):

donnie-jp (approved):
LGTM

Sample/ping.json
```diff
@@ -17,7 +17,7 @@
                     }
                 ],
                 "messagePayload": {
-                    "title": "[ToolTip] tooltip 1",
+                    "title": "[Tooltip] tooltip 1",
```
donnie-jp, line 20: @imaginaris did you change all occurrences of "ToolTip". Maybe we should have a mix of both forms in the unit tests?

imaginaris: I added a unit test for that (https://github.com/rakutentech/ios-inappmessaging/pull/210/files#diff-e2dbcd8583148f84ae18f400d37b5cbcaffaa5b013f33bf654d4a30c237728f7R137). Do you think it's worth having more than one form in json stubs as well?

malayli (approved):

donnie-jp (approved):

Sample/ping.json
donnie-jp: No - what you have now is sufficient. Thanks

PR #222, opened 2022-09-20
fix: fix incorrect status bar overlay color when app is in background (SDKCF-5134)

imaginaris:
# Description
In some apps, when app enters background, views are laid out again.
This makes the IAM SDK to reload status bar overlay view in displayed campaign.
Previously the code ignored non-foreground scenes when checking for current `statusBarStyle` which made statusbar overlay color to turn white when app re-enters foreground.

## Links
SDKCF-5134

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Extensions/UIApplication+IAM.swift
+++ b/Sources/RInAppMessaging/Extensions/UIApplication+IAM.swift
@@ -18,7 +18,7 @@ extension UIApplication {
     func getCurrentStatusBarStyle() -> UIStatusBarStyle? {
         if #available(iOS 13.0, *),
            let keyScene = connectedScenes
-                .filter({ $0.activationState == .foregroundActive })
+                .filter({ $0.activationState != .unattached })
                 .compactMap({ $0 as? UIWindowScene })
                 .first {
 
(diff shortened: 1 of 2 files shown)


imaginaris:
- [x] Do not merge before 7.2.0 release 
- [x] Update changelog after release

vincentdelmundo (approved):
LGTM

malayli (approved):

Sources/RInAppMessaging/Extensions/UIApplication+IAM.swift
```diff
@@ -18,7 +18,7 @@ extension UIApplication {
     func getCurrentStatusBarStyle() -> UIStatusBarStyle? {
         if #available(iOS 13.0, *),
            let keyScene = connectedScenes
-                .filter({ $0.activationState == .foregroundActive })
+                .filter({ $0.activationState != .unattached })
```
malayli, line 21: @imaginaris nit: In which case does it occur?
Is it a fix for apps having a SceneDelegate?

imaginaris: We shouldn't filter out the background activationState and any other state. I don't fully understand what exactly is a scene with `.unattached` activationState and I don't know how to trigger it. But it seemed like a state that we don't want to support (after looking at its description).

PR #234, opened 2022-11-21
build: fix unreachable config url in integration tests

imaginaris:
# Description
Fixed integration tests error: `Code=-1004 "Could not connect to the server."` during config endpoint request.
`RIAM_CONFIG_URL` override in Debug.xcconfig file is only useful for manual testing in Sample app.

## Links
https://app.bitrise.io/build/d6dc2c39-31f0-4a2f-a9f2-4681312d4562

# Checklist

### DIFF
--- a/Sample/Example-Debug.xcconfig
+++ b/Sample/Example-Debug.xcconfig
@@ -6,4 +6,4 @@
 // override secrets
 
 // uncomment below line to use mocked server
-RIAM_CONFIG_URL=http:/$()/localhost:6789/config
+// RIAM_CONFIG_URL=http:/$()/localhost:6789/config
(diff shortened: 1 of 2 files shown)


Sample/Example-Debug.xcconfig
```diff
@@ -6,4 +6,4 @@
 // override secrets
 
 // uncomment below line to use mocked server
```
malayli, line 8: Is the line `uncomment below line to use mocked server` still needed?

imaginaris: Yes. I left it for debugging purposes to easily mock endpoints and test campaigns display in the sample app.
This place is referenced here https://github.com/rakutentech/ios-inappmessaging/blob/21b6df2226d0bf472e4b22da0122cf36a75700fc/Sample/AppDelegate.swift#L17

malayli (approved):

PR #235, opened 2022-11-29
feat: add option to enable/disable Tooltip feature (SDKCF-6075)

imaginaris:
# Description
* Added a new flag in `configure()` API method to enable/disable the tooltip feature
* Refactored the way SDK configuration data is accessed in runtime

## Links
SDKCF-6075

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -273,6 +273,15 @@ The process can be considered as successful when `application(_:didRegisterForRe
 
 [How to set up your app for registering with APNS](https://developer.apple.com/documentation/usernotifications/registering_your_app_with_apns)
 
+### **Tooltip Campaigns**
+Tooltip feature is currently in beta testing; its features and behaviour might change in the future.
+Please refer to the internal guide for more information.
+
+To enable tooltips you must set `enableTooltipFeature` flag to true when calling `configure()`
+```swift
+RInAppMessaging.configure(enableTooltipFeature: true)
+```
+
 ## **Build/Run Example Application and Unit Tests**
 
 * Clone or fork the repo
--- a/Sources/RInAppMessaging/DependencyManager/MainContainer.swift
+++ b/Sources/RInAppMessaging/DependencyManager/MainContainer.swift
@@ -10,16 +10,12 @@ internal enum MainContainerFactory {
 
     private typealias ContainerElement = TypedDependencyManager.ContainerElement
 
-    static func create(dependencyManager manager: TypedDependencyManager, configURL: String?) -> TypedDependencyManager.Container {
+    static func create(dependencyManager manager: TypedDependencyManager, configURL: URL) -> TypedDependencyManager.Container {
 
         var elements = [
             ContainerElement(type: CommonUtility.self, factory: { CommonUtility() }),
             ContainerElement(type: ReachabilityType.self, factory: {
-                guard let configURL = URL(string: configURL ?? "") else {
-                    assertionFailure("Invalid Configuration URL: \(configURL ?? "<empty>")")
-                    return nil
-                }
-                return Reachability(url: configURL)
+                Reachability(url: configURL)
             }),
             ContainerElement(type: ConfigurationRepositoryType.self, factory: {
                 ConfigurationRepository()
--- a/Sources/RInAppMessaging/DependencyManager/MainContainer.swift
+++ b/Sources/RInAppMessaging/DependencyManager/MainContainer.swift
@@ -10,23 +10,13 @@ internal enum MainContainerFactory {
 
     private typealias ContainerElement = TypedDependencyManager.ContainerElement
 
-    private static func getValidConfigURL() -> URL? {
-        guard !Environment.isUnitTestEnvironment else {
-            return URL(string: "https://config.test")
-        }
-        guard let configURLString = BundleInfo.inAppConfigurationURL, !configURLString.isEmpty else {
-            return nil
-        }
-        return URL(string: configURLString)
-    }
-
-    static func create(dependencyManager manager: TypedDependencyManager) -> TypedDependencyManager.Container {
+    static func create(dependencyManager manager: TypedDependencyManager, configURL: String?) -> TypedDependencyManager.Container {
 
         var elements = [
             ContainerElement(type: CommonUtility.self, factory: { CommonUtility() }),
             ContainerElement(type: ReachabilityType.self, factory: {
-                guard let configURL = getValidConfigURL() else {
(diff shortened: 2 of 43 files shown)


README.md
```diff
@@ -273,6 +273,15 @@ The process can be considered as successful when `application(_:didRegisterForRe
 
 [How to set up your app for registering with APNS](https://developer.apple.com/documentation/usernotifications/registering_your_app_with_apns)
 
+### **Tooltip Campaigns**
+Tooltip feature is currently in beta testing; its features and behaviour might change in the future.
+Please refer to the internal guide for more information.
+
+To enable tooltips you must set `enableTooltipFeature` flag to true when calling `configure()`
+```swift
+RInAppMessaging.configure(enableTooltipFeature: true)
```
malayli, line 282: nit: Is it enabled by default?

Sources/RInAppMessaging/DependencyManager/MainContainer.swift
```diff
@@ -10,23 +10,13 @@ internal enum MainContainerFactory {
...
-            return URL(string: "https://config.test")
-        }
-        guard let configURLString = BundleInfo.inAppConfigurationURL, !configURLString.isEmpty else {
-            return nil
-        }
-        return URL(string: configURLString)
-    }
-
-    static func create(dependencyManager manager: TypedDependencyManager) -> TypedDependencyManager.Container {
+    static func create(dependencyManager manager: TypedDependencyManager, configURL: String?) -> TypedDependencyManager.Container {
```
malayli, line 14: If `configURL` is not necessary, can we have: `configURL: String? = nil`?

Sources/RInAppMessaging/DependencyManager/MainContainer.swift
```diff
@@ -10,23 +10,13 @@ internal enum MainContainerFactory {
...
 
         var elements = [
             ContainerElement(type: CommonUtility.self, factory: { CommonUtility() }),
             ContainerElement(type: ReachabilityType.self, factory: {
-                guard let configURL = getValidConfigURL() else {
-                    assertionFailure("Configuration URL in Info.plist is missing")
+                guard let configURL = URL(string: configURL ?? "") else {
```
malayli: Maybe clearer to have:
```
guard let url = configURL , let configURL = URL(string: url) else {
...
```

Sources/RInAppMessaging/InAppMessagingModuleConfiguration.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+internal struct InAppMessagingModuleConfiguration {
+    let configurationURL: String?
```
malayli: I'm wondering if we should call it `configurationURLString` as `configurationURL` seems to refer to `URL` type.

README.md
imaginaris: No, it is not.

Sources/RInAppMessaging/DependencyManager/MainContainer.swift
imaginaris: Yeah, I think we can

imaginaris: Actually I did some additional refactoring and moved checking validity of config url to `configure()` method. Please take a look at the new commit

Sources/RInAppMessaging/InAppMessagingModuleConfiguration.swift
imaginaris: Yeah. I think it's less confusing even though its longer and contains type's name

Sources/RInAppMessaging/RInAppMessaging.swift
```diff
@@ -53,24 +53,34 @@ import RSDKUtils
...
-    /// - Parameter subscriptionKey: your app's subscription key. (This setting will override the `InAppMessagingAppSubscriptionID` value in Info.plist)
-    /// - Parameter configurationURL: a configuration URL. (This setting will override the `InAppMessagingConfigurationURL` value in Info.plist)
+    /// - Parameters:
+    ///     - subscriptionKey: your app's subscription key. (This setting will override the `InAppMessagingAppSubscriptionID` value in Info.plist)
+    ///     - configurationURL: a configuration URL. (This setting will override the `InAppMessagingConfigurationURL` value in Info.plist)
+    ///     - enableTooltipFeature: set to `true` to enable Tooltip campaigns. This feature is currently in beta phase. Default value: `false`
     @objc public static func configure(subscriptionID: String? = nil,
-                                       configurationURL: String? = nil) {
+                                       configurationURL: String? = nil,
+                                       enableTooltipFeature: Bool = false) {
```
malayli, line 62: 👍

Sources/RInAppMessaging/Services/ConfigurationService.swift
```diff
@@ -12,20 +12,27 @@ internal enum ConfigurationServiceError: Error {
...
+    init(configurationRepository: ConfigurationRepositoryType) {
+        self.configurationRepository = configurationRepository
+        self.httpSession = URLSession(configuration: configurationRepository.defaultHttpSessionConfiguration)
     }
 
-    func getConfigData() -> Result<ConfigData, ConfigurationServiceError> {
+    func getConfigData() -> Result<ConfigEndpointData, ConfigurationServiceError> {
+        guard let configURLString = configurationRepository.getConfigEndpointURL(),
+              let configURL = URL(string: configURLString) else {
+            assertionFailure("Invalid Configuration URL: \(configurationRepository.getConfigEndpointURL() ?? "<empty>")")
```
malayli: Is `assertionFailure` necessary here as `.failure(.missingOrInvalidConfigURL)` is returned just after?

Sources/RInAppMessaging/Repositories/ConfigurationRepository.swift
```diff
@@ -46,6 +46,6 @@ internal class ConfigurationRepository: ConfigurationRepositoryType {
     }
 
     func getConfigEndpointURL() -> String? {
```
malayli: Maybe we should rename `getConfigEndpointURL` to `getConfigEndpointURLString`.
Otherwise `getConfigEndpointURL` can return `URL?` instead of `String?` if possible.

imaginaris: ok. Yeah I think it makes sense

malayli (approved):

PR #236, opened 2022-11-30
feat: track opt-in/opt-outs events in Push Primer feature (SDKCF-6026)

imaginaris:
# Description
Added tracking of opt-in/opt-out events in Push Primer feature.
The event are sent to RAnalytics whenever user allows or denies push notification authorization as a part of Push Primer feature.

## Links
SDKCF-6026

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 	- Added new API method to close displayed tooltips [SDKCF-6027]
 	- Added contexts validation for tooltip campaigns [SDKCF-6028]
 	- Added a feature flag to enable/disable the Tooltip feature [SDKCF-6075]
+	- Added tracking of opt-in/opt-outs in Push Primer feature [SDKCF-6026]
 - Bug fixes:
 	- Fixed an edge case of not readable status bar when campaign message is displayed [SDKCF-5134]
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Features:
 	- Added new API method to close displayed tooltips [SDKCF-6027]
 	- Added contexts validation for tooltip campaigns [SDKCF-6028]
+	- Added tracking of opt-in/opt-outs in Push Primer feature [SDKCF-6026]
```
vincentdelmundo, line 8: Please add a section or note on how to use the new API.

vincentdelmundo (changes requested):
LGTM. Please see inline comment

CHANGELOG.md
imaginaris: but it's not a new API. We don't have anything about analytics events sent by the SDK

Sources/RInAppMessaging/Views/Presenters/FullViewPresenter.swift
```diff
@@ -1,6 +1,12 @@
 import UIKit
 import UserNotifications
 
+#if canImport(RSDKUtils)
+import class RSDKUtils.AnalyticsBroadcaster
+#else // SPM Version
+import class RSDKUtilsMain.AnalyticsBroadcaster
+#endif
```
malayli, line 8: nit: it's only my opinion but it's maybe clearer to use `#if SWIFT_PACKAGE/#else/#endif`.

malayli (approved):

Sources/RInAppMessaging/Views/Presenters/FullViewPresenter.swift
imaginaris: Oh yeah that's a good suggestion. I agree. I will create another PR with that change in the whole module

vincentdelmundo (approved):

PR #243, opened 2022-12-16
improve: [Tooltip] fix UITabBar item support by adding new extension method (SDKCF-6134)

imaginaris:
# Description
Added new UITabBar extension method to provide a convenient way for setting accessibilityIdentifier in UITabBar buttons properly.

Note: Nimble version was synchronized with SPM version to avoid compilation errors (when using `allPass()` function)

## Links
SDKCF-6134

# Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,16 @@ To enable tooltips you must set `enableTooltipFeature` flag to true when calling
 RInAppMessaging.configure(enableTooltipFeature: true)
 ```
 
+#### **Displaying Tooltips on tab bar buttons**
+To be able to display tooltips on UITabBar items/buttons you need to set `accessibilityIdentifier` of UITabBarItem object associated with your tab view controller. Then you need to call `updateItemIdentifiers()` method on `UITabBar` object.
+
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -283,6 +283,16 @@ To enable tooltips you must set `enableTooltipFeature` flag to true when calling
 RInAppMessaging.configure(enableTooltipFeature: true)
 ```
 
+#### **Displaying Tooltips on tab bar buttons**
+To be able to display tooltips on UITabBar items/buttons you need to set `accessibilityIdentifier` of UITabBarItem object associated with your tab view controller. Then you need to call `updateItemIdentifiers()` method on `UITabBar` object.
+
+A code example from UIViewController's `viewDidLoad()` method
+```swift
+tabBarItem.accessibilityIdentifier = "UIElement.4"
```
malayli, line 291: Does it mean that only this `tabBarItem` instance will get a tooltip corresponding to `UIElement.4` identifier?

malayli (approved):

README.md
imaginaris: That's right

PR #251, opened 2023-03-13
fix: remove "last user" cache to keep anonymous user cache independent (SDKCF-6409)

imaginaris:
# Description
Removed the concept of "last user" cache to keep anonymous user cache independent and simplify the SDK's logic.
The anonymous user cache was overwritten with last user data on each init to provide continuity of data flow for the short time between moment when app is inited with empty user info and the moment when user preference object contains last logged in user info.
The SDK shouldn't try to predict login/logout/refresh behaviour in the host app.

## Links
SDKCF-6409

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## Changelog
 
+### Unreleased
+- Bug fixes:
+	- Removed "last user" cache to avoid overwriting anonymous user cache during init [SDKCF-6409]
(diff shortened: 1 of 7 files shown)


ts-dzmitry-kurski (approved):

ts-robih-robihamanto (approved):
LGTM

malayli (approved):

PR #257, opened 2023-03-23
tests: Add testing code to cover all code in the UserInfoProvider (SDKCF-6380)

ts-robih-robihamanto:
# Description
Increase code coverage in UserInfoProvider.swift
Include tests to cover `!=` when comparing the object.

## Links
SDKCF-6380

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/UserInfoProvider.swift
+++ b/Sources/RInAppMessaging/UserInfoProvider.swift
@@ -66,8 +66,14 @@ func == (lhs: UserInfoProvider?, rhs: UserInfoProvider?) -> Bool {
 
 @inlinable
 func != (lhs: UserInfoProvider?, rhs: UserInfoProvider?) -> Bool {
-    let equalsResult = lhs == rhs
-    return !equalsResult
+    switch (lhs, rhs) {
+    case let (l?, r?):
+        return l != r
+    case (nil, nil):
+        return false
+    default:
+        return true
+    }
 }
 
 @inlinable
--- a/Sources/RInAppMessaging/UserInfoProvider.swift
+++ b/Sources/RInAppMessaging/UserInfoProvider.swift
@@ -72,7 +72,7 @@ func != (lhs: UserInfoProvider?, rhs: UserInfoProvider?) -> Bool {
     case (nil, nil):
         return false
     default:
-        return true
+        return false
     }
 }
 
--- a/Tests/Tests/UserInfoProviderSpec.swift
+++ b/Tests/Tests/UserInfoProviderSpec.swift
@@ -157,6 +157,106 @@ class UserInfoProviderSpec: QuickSpec {
                     expect(userInfoProviderA == userInfoProviderB).to(beTrue())
                 }
             }
+
+            context("when checking inequality") {
+                var userInfoProviderA: UserInfoProviderMock!
+                var userInfoProviderB: UserInfoProviderMock!
+
+                beforeEach {
+                    userInfoProviderA = UserInfoProviderMock()
+                    userInfoProviderB = UserInfoProviderMock()
+                }
+
+                it("will return true for nil object") {
+                    userInfoProviderA = nil
+                    userInfoProviderB = nil
+                    expect(userInfoProviderA != userInfoProviderB).to(beFalse())
+                }
+
+                it("will return true for nil and empty objects") {
+                    userInfoProviderA = nil
+                    expect(userInfoProviderA != userInfoProviderB).to(beTrue())
+                }
+
+                it("will return false for different userIds") {
+                    userInfoProviderA.userID = "userA"
+                    userInfoProviderB.userID = "userA"
+                    expect(userInfoProviderA != userInfoProviderB).to(beFalse())
+                }
+
+                it("will return false for nil end empty userId") {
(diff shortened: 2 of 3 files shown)


Sources/RInAppMessaging/UserInfoProvider.swift
```diff
@@ -66,8 +66,14 @@ func == (lhs: UserInfoProvider?, rhs: UserInfoProvider?) -> Bool {
 
 @inlinable
 func != (lhs: UserInfoProvider?, rhs: UserInfoProvider?) -> Bool {
-    let equalsResult = lhs == rhs
-    return !equalsResult
+    switch (lhs, rhs) {
```
imaginaris, line 69: Was this change necessary?

Tests/Tests/UserInfoProviderSpec.swift
```diff
@@ -157,6 +157,106 @@ class UserInfoProviderSpec: QuickSpec {
...
+            context("when checking inequality") {
+                var userInfoProviderA: UserInfoProviderMock!
+                var userInfoProviderB: UserInfoProviderMock!
+
+                beforeEach {
+                    userInfoProviderA = UserInfoProviderMock()
+                    userInfoProviderB = UserInfoProviderMock()
+                }
+
+                it("will return true for nil object") {
```
imaginaris: I think you meant "will return false..."

Tests/Tests/UserInfoProviderSpec.swift
```diff
@@ -157,6 +157,106 @@ class UserInfoProviderSpec: QuickSpec {
...
+                    userInfoProviderB = nil
+                    expect(userInfoProviderA != userInfoProviderB).to(beFalse())
+                }
+
+                it("will return true for nil and empty objects") {
+                    userInfoProviderA = nil
+                    expect(userInfoProviderA != userInfoProviderB).to(beTrue())
+                }
+
+                it("will return false for different userIds") {
```
imaginaris, line 181: But userIds are not different in this test

imaginaris (changes requested):

Sources/RInAppMessaging/UserInfoProvider.swift
ts-robih-robihamanto: Yes, otherwise the code on 87 - 91 will not get called:
```
@inlinable
func != (lhs: UserInfoProvider, rhs: UserInfoProvider) -> Bool {
    let equalsResult = lhs == rhs
    return !equalsResult
}
```

Tests/Tests/UserInfoProviderSpec.swift
ts-robih-robihamanto: Oh ya, the user info is treated by userIds in this case should return false.

ts-robih-robihamanto: Fixed here [8b78146](https://github.com/rakutentech/ios-inappmessaging/pull/257/commits/a88a42758571ac69c74129b85cc4835ee8b78146).

Tests/Tests/UserInfoProviderSpec.swift
ts-robih-robihamanto: Fixed here [e4ab56e](https://github.com/rakutentech/ios-inappmessaging/pull/257/commits/e9822a7e2d95d60d8837db3f031f2b990e4ab56e).

Sources/RInAppMessaging/UserInfoProvider.swift
imaginaris: Hmm. But it would if you used non optional `UserInfoProviderMock` objects in your test, wouldn't it?

Tests/Tests/UserInfoProviderSpec.swift
```diff
@@ -157,6 +157,106 @@ class UserInfoProviderSpec: QuickSpec {
...
+
+                it("will return false for nil object") {
+                    userInfoProviderA = nil
+                    userInfoProviderB = nil
+                    expect(userInfoProviderA != userInfoProviderB).to(beFalse())
+                }
+
+                it("will return false for nil and empty objects") {
+                    userInfoProviderA = nil
+                    expect(userInfoProviderA != userInfoProviderB).to(beFalse())
```
imaginaris, line 178: so empty object and nil are treated as equal? 🤔 It doesn't sound correct to me

Sources/RInAppMessaging/UserInfoProvider.swift
imaginaris: but maybe it's better to keep it your way after all

Tests/Tests/UserInfoProviderSpec.swift
ts-robih-robihamanto: Yes, I think because both of the `userIds ` is empty it's equal.
In the UserInfoProvider | L.81 you made a comment:
**// null and empty values are treated the same**
So, it should be correct, right?

imaginaris: Ahh ok. That makes sense. I remember now 😅. Nil and empty user are treated as the same to improve 'logout' state detection. The test is absolutely correct.

imaginaris (approved):

PR #259, opened 2023-03-28
tests: Add testing code to cover all code in the EventMatcher (SDKCF-6390)

ts-robih-robihamanto:
# Description
Increase code coverage in EventMatcher.swift
Include tests to cover the failure case when matching the event.

## Links
SDKCF-6390

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/EventMatcher.swift
+++ b/Sources/RInAppMessaging/EventMatcher.swift
@@ -80,11 +80,10 @@ internal class EventMatcher: EventMatcherType {
         let events = matchedEvents.get()[campaign.id, default: []] + persistentEvents
         let allTriggersSatisfied = triggers.allSatisfy { isTriggerMatchingOneOfEvents($0, events: events) }
         
-        if campaign.isTooltip {
-            return allTriggersSatisfied && events.contains(where: { $0.type == .viewAppeared })
-        } else {
+        guard campaign.isTooltip else {
             return allTriggersSatisfied
         }
+        return allTriggersSatisfied && events.contains(where: { $0.type == .viewAppeared })
     }
 
     func removeSetOfMatchedEvents(_ eventsToRemove: Set<Event>, for campaign: Campaign) throws {
--- a/Tests/Tests/EventMatcherSpec.swift
+++ b/Tests/Tests/EventMatcherSpec.swift
@@ -75,8 +75,13 @@ class EventMatcherSpec: QuickSpec {
             context("when removing events") {
 
                 it("will throw error if events for given campaign weren't found") {
+                    let customEvent = CustomEvent(withName: "customEventTest", withCustomAttributes: nil)
+                    let customEventToRemove = CustomEvent(withName: "customEventToRemove", withCustomAttributes: nil)
+                    campaignRepository.list = [testCampaign]
+                    eventMatcher.matchAndStore(event: customEvent)
                     expect {
-                        try eventMatcher.removeSetOfMatchedEvents([AppStartEvent()], for: testCampaign)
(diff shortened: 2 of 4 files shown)


Sources/RInAppMessaging/EventMatcher.swift
```diff
@@ -80,11 +80,10 @@ internal class EventMatcher: EventMatcherType {
         let events = matchedEvents.get()[campaign.id, default: []] + persistentEvents
         let allTriggersSatisfied = triggers.allSatisfy { isTriggerMatchingOneOfEvents($0, events: events) }
         
-        if campaign.isTooltip {
-            return allTriggersSatisfied && events.contains(where: { $0.type == .viewAppeared })
-        } else {
+        guard campaign.isTooltip else {
```
malayli, line 83: @ts-robih-robihamanto Does this change improve the code coverage?
Or is it an improvement for readability by using `guard`?

Tests/Tests/EventMatcherSpec.swift
```diff
@@ -85,6 +89,17 @@ class EventMatcherSpec: QuickSpec {
                     }.to(throwError(EventMatcherError.couldntFindRequestedSetOfEvents))
                 }
 
+                it("will throw error if the event to remove isn't persistent or the event weren't found inside persistent events") {
```
imaginaris: 1. "... the event wasn't"
2. "or" suggests this test should be split into 2 tests to cover all scenarios
3. CustomEvent can't be a presistent event. I don't understand how persistent events are used in this test case 🤔

Tests/Tests/EventMatcherSpec.swift
```diff
@@ -212,6 +227,16 @@ class EventMatcherSpec: QuickSpec {
                 expect(events).to(contain(AppStartEvent()))
             }
 
+            it("will stop matching events when triggers inside campaign is nil") {
```
imaginaris: I'm not sure 'stop' is a good verb here. How about "will not match any event with a campaign that has no triggers"

imaginaris (commented):
@ts-robih-robihamanto Please write in the PR description the new code coverage for EventMatcher

Sources/RInAppMessaging/EventMatcher.swift
ts-robih-robihamanto: @malayli 
Yes, It does improve the code coverage, previously the line after else {} is never executed, even though only empty space. But, Xcode shows it as not a covered line of code.

Tests/Tests/EventMatcherSpec.swift
ts-robih-robihamanto: Good point, I will change the description.

Tests/Tests/EventMatcherSpec.swift
ts-robih-robihamanto: 1. Yes, I change this.
2. I change the description to only one because I think the conditional is to check whether the event is a persistent event or not.
3. That's why we use this case to cover line 106 in EventMatcher to return an `error:EventMatcherError.couldntFindRequestedSetOfEvents` because the CustomEvent is not a persistent one.

ts-dzmitry-kurski (approved):

Tests/Tests/EventMatcherSpec.swift
imaginaris: Ok, I see. The name is still a bit confusing because the EventMatcher is able to remove an event that isn't persistent without throwing an error. Actually, please move content of this test to "will throw error if events for given campaign weren't found" test and replace its content. The replaced content should be put under a new test with name "will throw error if persistent event for given campaign wasn't found"

imaginaris (approved):

PR #260, opened 2023-03-29
tests: improved the unit test code coverage to SlideUpViewPresenter.swift (SDKCF-6383)

Esakkiraja-Pothikannan:
# Description
- Added `slide-up-trigger-without-messageBody.json` file with an invalid response
- Added coverage to `guard else` statement in `viewDidInitialize()`
- Removed unused `completionHandler` 

## Links
- SDKCF-6383

# Checklist
- [ ] I removed all sensitive data **before every commit**, including API endpoints and keys, and internal links

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,11 @@
 	- Added display permission service to perform ping on tooltip dispatcher [SDKCF-4964]
 - Improvements:
 	- Created new test scheme to combine unit test and UI test results [SDKCF-6356]
-	- Added a new file for EventType unit test to increase code coverage [SDKCF-6378]
-	- Added more unit test to increase code coverage [SDKCF-6379]
-	- Added testing code to cover all code in the UserInfoProvider [SDKCF-6380]
+	- Improved unit test code coverage to
+		- EventType.swift [SDKCF-6378]
+		- CustomAttribute.swift [SDKCF-6379]
+		- UserInfoProvider.swift [SDKCF-6380]
+		- SlideUpViewPresenter.swift [SDKCF-6383]
 
 ### 7.3.0 (2023-01-11)
 - Features:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 		- CustomAttribute.swift [SDKCF-6379]
 		- UserInfoProvider.swift [SDKCF-6380]
 		- EventMatcher.swift [SDKCF-6390]
+		- ConfigurationService.swift [SDKCF-6388]
 		- SlideUpViewPresenter.swift [SDKCF-6383]
 
 ### 7.3.0 (2023-01-11)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 		- EventType.swift [SDKCF-6378]
 		- CustomAttribute.swift [SDKCF-6379]
 		- UserInfoProvider.swift [SDKCF-6380]
+		- EventMatcher.swift [SDKCF-6390]
 		- SlideUpViewPresenter.swift [SDKCF-6383]
 
 ### 7.3.0 (2023-01-11)
--- a/RInAppMessaging.xcodeproj/project.pbxproj
+++ b/RInAppMessaging.xcodeproj/project.pbxproj
@@ -73,6 +73,7 @@
 		45BDFD032421D311004DEA0C /* CampaignsListManagerSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45BDFD022421D311004DEA0C /* CampaignsListManagerSpec.swift */; };
 		45BDFD052421EE7B004DEA0C /* SharedMocks.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45BDFD042421EE7B004DEA0C /* SharedMocks.swift */; };
 		45BDFD0724232714004DEA0C /* PublicAPISpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45BDFD0624232714004DEA0C /* PublicAPISpec.swift */; };
+		45D064D029D2ECE10009DFB2 /* slide-up-trigger-without-messageBody.json in Resources */ = {isa = PBXBuildFile; fileRef = 45D064CE29D2ECE10009DFB2 /* slide-up-trigger-without-messageBody.json */; };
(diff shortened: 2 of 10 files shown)


ts-robih-robihamanto:
@imaginaris 
Do we need to add a new file in this case `slide-up-trigger-without-messageBody.json` to `SampleSPM/SampleSPM.xcodeproj`? or it's just for the Spec file?

Esakkiraja-Pothikannan:
> @imaginaris Do we need to add a new file in this case `slide-up-trigger-without-messageBody.json` to `SampleSPM/SampleSPM.xcodeproj`? or it's just for the Spec file?

This is only for `UITests` target.

CHANGELOG.md
```diff
@@ -6,9 +6,11 @@
 	- Added display permission service to perform ping on tooltip dispatcher [SDKCF-4964]
 - Improvements:
 	- Created new test scheme to combine unit test and UI test results [SDKCF-6356]
-	- Added a new file for EventType unit test to increase code coverage [SDKCF-6378]
-	- Added more unit test to increase code coverage [SDKCF-6379]
-	- Added testing code to cover all code in the UserInfoProvider [SDKCF-6380]
+	- Improved unit test code coverage to
```
imaginaris, line 9: 👍

imaginaris:
> @imaginaris
> Do we need to add a new file in this case `slide-up-trigger-without-messageBody.json` to `SampleSPM/SampleSPM.xcodeproj`? or it's just for the Spec file?

That's right. All new files, including JSON files, should be also added to `SampleSPM.xcodeproj`

RInAppMessaging.xcodeproj/project.pbxproj
```diff
@@ -73,6 +73,7 @@
 		45BDFD032421D311004DEA0C /* CampaignsListManagerSpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45BDFD022421D311004DEA0C /* CampaignsListManagerSpec.swift */; };
 		45BDFD052421EE7B004DEA0C /* SharedMocks.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45BDFD042421EE7B004DEA0C /* SharedMocks.swift */; };
 		45BDFD0724232714004DEA0C /* PublicAPISpec.swift in Sources */ = {isa = PBXBuildFile; fileRef = 45BDFD0624232714004DEA0C /* PublicAPISpec.swift */; };
+		45D064D029D2ECE10009DFB2 /* slide-up-trigger-without-messageBody.json in Resources */ = {isa = PBXBuildFile; fileRef = 45D064CE29D2ECE10009DFB2 /* slide-up-trigger-without-messageBody.json */; };
```
Esakkiraja-Pothikannan, line 76: @imaginaris Why test target file should be added to `SampleSPM.xcodeproj` ?

Tests/UITests/SlideUpViewSpec.swift
```diff
@@ -40,6 +40,20 @@ class SlideUpViewSpec: QuickSpec {
 
         describe("Slide-up campaign view") {
 
+            context("trigger action with invalid message payload") {
+
+                beforeEach {
+                    launchAppIfNecessary(context: "slide-up-trigger-without-messageBody")
+                    if !iamView.exists {
```
malayli: @Esakkiraja-Pothikannan Why is this condition needed?

RInAppMessaging.xcodeproj/project.pbxproj
imaginaris: Because we use `SampleSPM.xcodeproj` to run tests in SPM mode. The CI will attempt to build 'Tests' target in `SampleSPM.xcodeproj`. So it's better to keep that project in sync with the original RInAppMessaging.xcodeproj

Tests/UITests/SlideUpViewSpec.swift
imaginaris: Yeah, In this case instead of this condition you should check if iamView is not presented: `expect(iamView.exists).to(beFalse())`

Tests/UITests/SlideUpViewSpec.swift
```diff
@@ -40,6 +40,20 @@ class SlideUpViewSpec: QuickSpec {
...
+                        app.buttons["purchase_successful"].tap()
+                    }
+                }
+
+                it("should not construct the slideup view") {
+                    expect(iamView.exists).to(beFalse())
```
imaginaris: It might take some time for IAM decide if `iamView` should be displayed or not. So you need to use `.toAfterTimeout(beFalse(), timeout: 2)` predicate.

Tests/UITests/SlideUpViewSpec.swift
Esakkiraja-Pothikannan: I don't think, it makes any difference from the current implementation. so it's fine with `if condition` right ?

imaginaris: This condition just does not match the expected behaviour. In other contexts it ensures the iamView is present before running any tests cases. In this context you don't want that to happen. So just having `app.buttons["purchase_successful"].tap()` should do

Esakkiraja-Pothikannan: ok got it, I understood other contexts `if condition` in a different way. So `expect(iamView.exists).to(beFalse())` and if `!iamView.exists {` is not required, we can directly call `app.buttons["purchase_successful"].tap()`. Will update this.

Sources/RInAppMessaging/EventMatcher.swift
```diff
@@ -80,11 +80,10 @@ internal class EventMatcher: EventMatcherType {
         let events = matchedEvents.get()[campaign.id, default: []] + persistentEvents
         let allTriggersSatisfied = triggers.allSatisfy { isTriggerMatchingOneOfEvents($0, events: events) }
         
-        if campaign.isTooltip {
-            return allTriggersSatisfied && events.contains(where: { $0.type == .viewAppeared })
-        } else {
+        guard campaign.isTooltip else {
```
malayli, line 83: @Esakkiraja-Pothikannan Was `guard` needed for coverage here?
Or you made this change because it is more readable than if else?

imaginaris (approved):

Esakkiraja-Pothikannan:
closing this as [PR-266](https://github.com/rakutentech/ios-inappmessaging/pull/266) has all the changes

PR #261, opened 2023-03-30
tests: improve Code Coverage of ConfigurationService (SDKCF-6388)

ts-robih-robihamanto:
# Description
Improved coverage of `ConfigurationService` to **98.9%**
This else line cannot failed, because it's already checked in the previous line if it's link is a valid URL:
```swift
guard let url = urlComponents?.url else {
    return .failure(RequestError.urlIsNil)
}
```

## Links
SDKCF-6388

# Checklist

### DIFF
--- a/Tests/Tests/ConfigurationServiceSpec.swift
+++ b/Tests/Tests/ConfigurationServiceSpec.swift
@@ -254,6 +254,15 @@ class ConfigurationServiceSpec: QuickSpec {
                 }
             }
 
+            context("when building request body") {
+
+                it("will return RequestError.bodyIsNil") {
+                    let result = service.buildHttpBody(with: nil)
+                    let error = result.getError() as? RequestError
+                    expect(error).toNot(beNil())
+                }
+            }
+
             context("when making a request") {
                 beforeEach {
                     BundleInfoMock.reset()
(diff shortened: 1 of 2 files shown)


Tests/Tests/ConfigurationServiceSpec.swift
```diff
@@ -254,6 +254,15 @@ class ConfigurationServiceSpec: QuickSpec {
                 }
             }
 
+            context("when building request body") {
+
+                it("will return RequestError.bodyIsNil") {
```
imaginaris: please add test cases that check if the same is returned when you pass a non-nil parameter in `buildHttpBody()` so it's clear that `bodyIsNil` is always returned no matter what parameter is passed

imaginaris (changes requested):
The `assertionFailure()` in `buildRequestHeader()` can be tested by setting subscription id to nil and using `to(throwAssertion())` predicate. It won't crash the unit test run

ts-robih-robihamanto:
> The `assertionFailure()` in `buildRequestHeader()` can be tested by setting subscription id to nil and using `to(throwAssertion())` predicate. It won't crash the unit test run

I added the case for throwAssertion here [0726cfa](https://github.com/rakutentech/ios-inappmessaging/pull/261/commits/62da5bc73d4b56f83df4c5bd4f4234dd20726cfa).

imaginaris (approved):
👍

PR #265, opened 2023-04-05
tests: add new testing file for ui label extension (SDKCF-6385)

ts-robih-robihamanto:
# Description
add a new testing file for `UILabel+IAM` extension covering 100% code coverage for this file.

## Links
SDKCF-6385

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Extensions/UILabel+IAM.swift
+++ b/Sources/RInAppMessaging/Extensions/UILabel+IAM.swift
@@ -6,7 +6,7 @@ internal extension UILabel {
     /// - Parameter lineSpacing: The value of the spacing for each line. Defaults to 0.
     func setLineSpacing(lineSpacing: CGFloat = 0.0) {
 
-        guard let labelText = self.text else {
+        guard let attributedText = self.attributedText else {
             return
         }
 
@@ -14,11 +14,7 @@ internal extension UILabel {
         paragraphStyle.lineSpacing = lineSpacing
 
         let attributedString: NSMutableAttributedString
-        if let attributedText, attributedText.length > 0 {
-            attributedString = NSMutableAttributedString(attributedString: attributedText)
-        } else {
-            attributedString = NSMutableAttributedString(string: labelText)
-        }
+        attributedString = NSMutableAttributedString(attributedString: attributedText)
 
         attributedString.addAttribute(NSAttributedString.Key.paragraphStyle,
                                       value: paragraphStyle,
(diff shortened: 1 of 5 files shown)


Sources/RInAppMessaging/Extensions/UILabel+IAM.swift
```diff
@@ -14,7 +14,7 @@ internal extension UILabel {
         paragraphStyle.lineSpacing = lineSpacing
 
         let attributedString: NSMutableAttributedString
-        if let attributedText = self.attributedText {
+        if let attributedText, attributedText.length > 0 {
```
ts-robih-robihamanto: Adding new condition because attributedText will never be nil.
source: [link](https://developer.apple.com/forums/thread/118581)

Esakkiraja-Pothikannan: `attributedText` would be nil if `UILabel().text` is nil. Since already `guard else` check is there at the top(line 9), this `if` condition will pass all the time in the both old and new approaches. Better, have a look and remove `else` part.

imaginaris: When I'm creating an instance of `Label()` I get `nil` for `attributedText`. It's true that if you set `text` to some non-nil value, you'll also get a non-nil value from `attributedText`. Setting `text` to nil also sets `attributedText` to nil. So I don't see any reason to add that additional condition.

imaginaris: You can replace the guard condition to `guard let attributedText = self.attributedText` and then proceed without any additional conditions and always use `NSMutableAttributedString(attributedString: attributedText)`

imaginaris: And also I think we don't have to restrict this function to non-empty texts only

imaginaris (changes requested):

Sources/RInAppMessaging/Extensions/UILabel+IAM.swift
```diff
@@ -6,19 +6,15 @@ internal extension UILabel {
     /// - Parameter lineSpacing: The value of the spacing for each line. Defaults to 0.
     func setLineSpacing(lineSpacing: CGFloat = 0.0) {
 
-        guard let labelText = self.text else {
+        guard let _ = text, let attributedText else {
```
imaginaris: 1. `let _ = text` this part is not needed
2. `let attributedText` is correct but, if I'm not mistaken, swift version available in Xcode 12.5 (which we still support) doesn't understand this shortened unwrapping yet so it's better to have `let attributedText = attributedText`

imaginaris (approved):

PR #273, opened 2023-04-20
tests: improve code coverage for http requestable (SDKCF-6389)

ts-robih-robihamanto:
# Description
Improve HttpRequestable file code coverage to 97.9%.

This code will never be executed because the completion handler is always called within `requestFromServer` function, considering removing this part will cover this file to 100% code coverage.
```swift
        guard let unwrappedResult = result else {
            Logger.debug("Error: Didn't get any result - completion handler not called!")
            assertionFailure()
            return .failure(.unknown)
        }
```

## Links
SDKCF-6389

# Checklist

### DIFF
--- a/Tests/Tests/HttpRequestableSpec.swift
+++ b/Tests/Tests/HttpRequestableSpec.swift
@@ -26,7 +26,10 @@ class HttpRequestableSpec: QuickSpec {
             }
 
             afterEach {
-                httpRequestable = nil // force deallocation of .httpSessionMock
+                // force deallocation of httpSessionMock
+                httpRequestable.httpSessionMock.responseData = nil
+                httpRequestable.httpSessionMock.responseError = nil
+                httpRequestable.httpSessionMock.sentRequest = nil
             }
 
             context("when calling requestFromServerSync") {
@@ -348,9 +351,9 @@ class HttpRequestableSpec: QuickSpec {
                     }
                 }
 
-                it("will whrow assertion if calling the method on the main thread") {
+                it("will throw assertion if calling the method on the main thread") {
                     expect(httpRequestable.requestFromServerSync(
-                        url: URL(string: "")!,
+                        url: URL(string: "https://test.url")!,
                         httpMethod: .get,
                         parameters: nil,
                         addtionalHeaders: nil)).to(throwAssertion())
--- a/Tests/Tests/HttpRequestableSpec.swift
+++ b/Tests/Tests/HttpRequestableSpec.swift
@@ -320,6 +320,41 @@ class HttpRequestableSpec: QuickSpec {
                         }
                     }
                 }
+
+                it("will using empty string for error message as a default when failed to encode error message as .utf8") {
+                    let data = "テスト".data(using: .japaneseEUC)!
+                    let response = HTTPURLResponse(
+                        url: URL(string: "https://test.url")!,
+                        statusCode: 404,
+                        httpVersion: nil,
+                        headerFields: nil)
+
+                    httpRequestable.httpSessionMock.responseData = data
+                    httpRequestable.httpSessionMock.httpResponse = response
+
+                    waitUntil { done in
+                        requestQueue.async {
+                            let result = httpRequestable.requestFromServerSync(
+                                url: URL(string: "https://test.url")!,
+                                httpMethod: .put,
+                                parameters: nil,
+                                addtionalHeaders: nil)
+
+                            expect {
+                                try result.get()
+                            }.to(throwError())
+                            done()
+                        }
+                    }
+                }
+
+                it("will whrow assertion if calling the method on the main thread") {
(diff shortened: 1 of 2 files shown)


Tests/Tests/HttpRequestableSpec.swift
```diff
@@ -320,6 +320,41 @@ class HttpRequestableSpec: QuickSpec {
...
+                        httpVersion: nil,
+                        headerFields: nil)
+
+                    httpRequestable.httpSessionMock.responseData = data
+                    httpRequestable.httpSessionMock.httpResponse = response
+
+                    waitUntil { done in
+                        requestQueue.async {
+                            let result = httpRequestable.requestFromServerSync(
+                                url: URL(string: "https://test.url")!,
```
malayli: @ts-robih-robihamanto This url is repeated twice.
Can you declare it once as `let url: URL! = URL(string: "https://test.url")` then use `url` here and at line 327?
Thank you.

Tests/Tests/HttpRequestableSpec.swift
```diff
@@ -320,6 +320,41 @@ class HttpRequestableSpec: QuickSpec {
...
+
+                            expect {
+                                try result.get()
+                            }.to(throwError())
+                            done()
+                        }
+                    }
+                }
+
+                it("will whrow assertion if calling the method on the main thread") {
```
imaginaris: "whrow" -> "throw"

Tests/Tests/HttpRequestableSpec.swift
```diff
@@ -320,6 +320,41 @@ class HttpRequestableSpec: QuickSpec {
...
+                    expect(httpRequestable.requestFromServerSync(
+                        url: URL(string: "")!,
```
imaginaris: `URL(string: "")!` will crash and falsify the test result. please use anything but an empty string in the URL constructor.

Tests/Tests/HttpRequestableSpec.swift
ts-robih-robihamanto: @malayli 
This URL `https://test.url` is used by 35 cases in this file, do I need to make declare it once and use the URL for all of the test cases?

ts-robih-robihamanto:
This code will never be executed because the completion handler is always called within `requestFromServer` function, considering removing this part will cover this file to 100% code coverage.
```swift
        guard let unwrappedResult = result else {
            Logger.debug("Error: Didn't get any result - completion handler not called!")
            assertionFailure()
            return .failure(.unknown)
        }
```

Tests/Tests/HttpRequestableSpec.swift
imaginaris: @ts-robih-robihamanto I think it makes sense. One constant for the whole file

Tests/Tests/HttpRequestableSpec.swift
```diff
@@ -26,7 +26,10 @@ class HttpRequestableSpec: QuickSpec {
             }
 
             afterEach {
-                httpRequestable = nil // force deallocation of .httpSessionMock
+                // force deallocation of httpSessionMock
```
imaginaris: We are not forcing deallocation here. Just simply clearing the state of `httpRequestable`. I think you can just remove this comment

Tests/Tests/HttpRequestableSpec.swift
```diff
@@ -320,6 +323,41 @@ class HttpRequestableSpec: QuickSpec {
...
+                            expect {
+                                try result.get()
+                            }.to(throwError())
+                            done()
+                        }
+                    }
+                }
+
+                it("will throw assertion if calling the method on the main thread") {
+                    expect(httpRequestable.requestFromServerSync(
```
imaginaris, line 356: Please add a comment here that `throwAssertion()` captures a strong reference to `httpRequestable` and never releases it making it impossible to deallocate the `httpSessionMock` reusable instance

ts-dzmitry-kurski (approved):

Tests/Tests/HttpRequestableSpec.swift
ts-robih-robihamanto: @malayli cc @imaginaris 
Added on this [be853e2](https://github.com/rakutentech/ios-inappmessaging/pull/273/commits/be853e2b2b0b107e942a87a69858c3b40da557ec) commit.

Tests/Tests/HttpRequestableSpec.swift
ts-robih-robihamanto: Added here [be853e2](https://github.com/rakutentech/ios-inappmessaging/pull/273/commits/be853e2b2b0b107e942a87a69858c3b40da557ec).

imaginaris: Thanks

imaginaris (approved):

PR #275, opened 2023-04-26
tests: improve code coverage for Router (SDKCF-6382)

ts-robih-robihamanto:
# Description
- Improve Router file code coverage from 89.3 to **97.1%**.
- Doing some refactoring to make the code more testable and I think it needs a QA test to make sure Campaign and Tooltip display correctly as in the previous behavior.

## Links
SDKCF-6382

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Models/Responses/Campaign.swift
+++ b/Sources/RInAppMessaging/Models/Responses/Campaign.swift
@@ -52,10 +52,10 @@ internal struct Campaign: Codable, Hashable {
         try container.encode(isOptedOut, forKey: .isOptedOut)
     }
 
-    init(data: CampaignData, asTooltip isTooltip: Bool = true) {
+    init(data: CampaignData) {
         self.data = data
         impressionsLeft = data.maxImpressions
-        tooltipData = isTooltip ? Campaign.parseTooltipData(messagePayload: data.messagePayload) : nil
+        tooltipData = Campaign.parseTooltipData(messagePayload: data.messagePayload)
     }
 
     static func == (lhs: Campaign, rhs: Campaign) -> Bool {
--- a/Sources/RInAppMessaging/Models/Responses/Campaign.swift
+++ b/Sources/RInAppMessaging/Models/Responses/Campaign.swift
@@ -52,10 +52,10 @@ internal struct Campaign: Codable, Hashable {
         try container.encode(isOptedOut, forKey: .isOptedOut)
     }
 
-    init(data: CampaignData) {
+    init(data: CampaignData, asTooltip isTooltip: Bool = true) {
         self.data = data
         impressionsLeft = data.maxImpressions
-        tooltipData = Campaign.parseTooltipData(messagePayload: data.messagePayload)
+        tooltipData = isTooltip ? Campaign.parseTooltipData(messagePayload: data.messagePayload) : nil
     }
 
     static func == (lhs: Campaign, rhs: Campaign) -> Bool {
--- a/Sources/RInAppMessaging/Router.swift
+++ b/Sources/RInAppMessaging/Router.swift
@@ -43,7 +43,7 @@ internal protocol RouterType: ErrorReportable {
 /// Handles all the displaying logic of the SDK.
 internal class Router: RouterType, ViewListenerObserver {
 
-    private enum UIConstants {
+    enum UIConstants {
         enum Tooltip {
             static let minDistFromEdge: CGFloat = 20.0
             static let targetViewSpacing: CGFloat = 0.0
@@ -313,7 +313,6 @@ internal class Router: RouterType, ViewListenerObserver {
             guard accessibilityCompatibleDisplay,
                let transitionViewClass = NSClassFromString("UITransitionView"),
                let transitionView = superview.subviews.first(where: { !$0.isKind(of: transitionViewClass) }) else {
-                
                 return superview
             }
             return transitionView
@@ -332,9 +331,6 @@ internal class Router: RouterType, ViewListenerObserver {
     }
 
     private func updateFrame(targetView: UIView, tooltipView: TooltipView, superview: UIView, position: TooltipBodyData.Position) {
-        guard targetView.superview != nil else {
-            return
-        }
         let targetViewFrame = superview.convert(targetView.frame, from: targetView.superview)
 
         switch position {
--- a/Sources/RInAppMessaging/Router.swift
+++ b/Sources/RInAppMessaging/Router.swift
@@ -43,7 +43,7 @@ internal protocol RouterType: ErrorReportable {
 /// Handles all the displaying logic of the SDK.
 internal class Router: RouterType, ViewListenerObserver {
 
-    enum UIConstants {
+    private enum UIConstants {
         enum Tooltip {
             static let minDistFromEdge: CGFloat = 20.0
             static let targetViewSpacing: CGFloat = 0.0
@@ -166,19 +166,20 @@ internal class Router: RouterType, ViewListenerObserver {
         var view: (() -> BaseView)!
         let type = campaign.data.type
 
-        if type == .modal || type == .full {
+        switch type {
+        case .modal, .full:
             let presenter = presenter as! FullViewPresenterType
             presenter.campaign = campaign
             if let associatedImageData = associatedImageData {
                 presenter.associatedImage = UIImage(data: associatedImageData)
             }
             view = type == .modal ? { ModalView(presenter: presenter) } : { FullScreenView(presenter: presenter) }
-        }
-
-        if type == .slide {
+        case .slide:
             let presenter = presenter as! SlideUpViewPresenterType
             presenter.campaign = campaign
             view = { SlideUpView(presenter: presenter) }
+        case .invalid, .html:
+            Logger.debug("Error: Campaign view type not supported")
         }
         return view
     }
--- a/Sources/RInAppMessaging/Router.swift
+++ b/Sources/RInAppMessaging/Router.swift
@@ -91,11 +91,10 @@ internal class Router: RouterType, ViewListenerObserver {
         let result = {
             return self.displayedTooltips[uiElementIdentifier] != nil
         }
-        if Thread.current == .main {
-            return result()
-        } else {
+        guard Thread.current == .main else {
             return DispatchQueue.main.sync(execute: result)
         }
+        return result()
     }
 
     func displayCampaign(_ campaign: Campaign,
@@ -293,14 +292,13 @@ internal class Router: RouterType, ViewListenerObserver {
         // For accessibilityCompatible option, campaign view must be inserted to
         // UIWindow's main subview. Private instance of UITransitionView
         // shouldn't be used for that - that's why it's omitted.
-        if accessibilityCompatibleDisplay,
+        
+        guard accessibilityCompatibleDisplay,
            let transitionViewClass = NSClassFromString("UITransitionView"),
-           let mainSubview = rootView.subviews.first(where: { !$0.isKind(of: transitionViewClass) }) {
-
-            return mainSubview
-        } else {
+           let mainSubview = rootView.subviews.first(where: { !$0.isKind(of: transitionViewClass) }) else {
             return rootView
         }
+        return mainSubview
     }
 
     private func findParentViewForTooltip(_ sourceView: UIView) -> UIView {
@@ -312,14 +310,13 @@ internal class Router: RouterType, ViewListenerObserver {
         case is UIScrollView:
             return superview
         case is UIWindow:
-            if accessibilityCompatibleDisplay,
+            guard accessibilityCompatibleDisplay,
                let transitionViewClass = NSClassFromString("UITransitionView"),
-               let transitionView = superview.subviews.first(where: { !$0.isKind(of: transitionViewClass) }) {
-
-                return transitionView
-            } else {
+               let transitionView = superview.subviews.first(where: { !$0.isKind(of: transitionViewClass) }) else {
+                
                 return superview
             }
+            return transitionView
         default:
             return findParentViewForTooltip(superview)
         }
--- a/Sources/RInAppMessaging/Router.swift
+++ b/Sources/RInAppMessaging/Router.swift
@@ -109,11 +109,15 @@ internal class Router: RouterType, ViewListenerObserver {
             return
         }
 
+        guard let presenter = getPresenter(for: campaignViewType) else {
+            Logger.debug("Error: \(campaignViewType) couldn't be resolved")
+            assertionFailure()
+            completion(true)
+            return
+        }
+
         displayQueue.async {
-            guard let viewConstructor = self.createViewConstructor(for: campaign, associatedImageData: associatedImageData) else {
-                completion(true)
-                return
-            }
+            let viewConstructor = self.createViewConstructor(for: campaign, presenter: presenter, associatedImageData: associatedImageData)
 
             DispatchQueue.main.async {
                 guard let rootView = UIApplication.shared.getKeyWindow(),
@@ -131,46 +135,52 @@ internal class Router: RouterType, ViewListenerObserver {
         }
     }
 
-    private func createViewConstructor(for campaign: Campaign, associatedImageData: Data?) -> (() -> BaseView)? {
-        func getPresenter<T>(type: T.Type) -> T? {
-            guard let presenter = self.dependencyManager.resolve(type: type) else {
-                Logger.debug("Error: \(type) couldn't be resolved")
-                assertionFailure()
+    private func getPresenter(for type: CampaignDisplayType) -> BaseViewPresenterType? {
+        var presenter: BaseViewPresenterType?
+        switch type {
+        case .modal, .full:
+            guard let resolvedPresenter = resolvePresenter(type: FullViewPresenterType.self) else {
                 return nil
             }
-            return presenter
-        }
-
-        switch campaign.data.type {
-        case .modal:
-            guard let presenter = getPresenter(type: FullViewPresenterType.self) else {
-                break
-            }
-            presenter.campaign = campaign
-            if let associatedImageData = associatedImageData {
-                presenter.associatedImage = UIImage(data: associatedImageData)
-            }
-            return { ModalView(presenter: presenter) }
-        case .full:
-            guard let presenter = getPresenter(type: FullViewPresenterType.self) else {
-                break
-            }
-            presenter.campaign = campaign
-            if let associatedImageData = associatedImageData {
-                presenter.associatedImage = UIImage(data: associatedImageData)
-            }
-            return { FullScreenView(presenter: presenter) }
+            presenter = resolvedPresenter
         case .slide:
-            guard let presenter = getPresenter(type: SlideUpViewPresenterType.self) else {
-                break
+            guard let resolvedPresenter = resolvePresenter(type: SlideUpViewPresenterType.self) else {
+                return nil
             }
+            presenter = resolvedPresenter
+        default:
+            Logger.debug("Error: Campaign view type is not supported")
+        }
+        return presenter
+    }
+
+    private func resolvePresenter<T>(type: T.Type) -> T? {
+        guard let presenter = self.dependencyManager.resolve(type: type) else {
+            Logger.debug("Error: \(type) couldn't be resolved")
+            return nil
+        }
+        return presenter
+    }
+
+    private func createViewConstructor(for campaign: Campaign, presenter: BaseViewPresenterType, associatedImageData: Data?) -> (() -> BaseView) {
+        var view: (() -> BaseView)!
+        let type = campaign.data.type
+
+        if type == .modal || type == .full {
+            let presenter = presenter as! FullViewPresenterType
             presenter.campaign = campaign
-            return { SlideUpView(presenter: presenter) }
-        case .invalid, .html:
-            Logger.debug("Error: Campaign view type not supported")
+            if let associatedImageData = associatedImageData {
+                presenter.associatedImage = UIImage(data: associatedImageData)
+            }
+            view = type == .modal ? { ModalView(presenter: presenter) } : { FullScreenView(presenter: presenter) }
         }
 
-        return nil
+        if type == .slide {
+            let presenter = presenter as! SlideUpViewPresenterType
+            presenter.campaign = campaign
+            view = { SlideUpView(presenter: presenter) }
+        }
+        return view
     }
 
     // swiftlint:disable:next function_parameter_count
--- a/Tests/Tests/Helpers/TestHelpers.swift
+++ b/Tests/Tests/Helpers/TestHelpers.swift
@@ -84,10 +84,7 @@ struct TestHelpers {
                                 position: TooltipBodyData.Position = .topCenter,
                                 messageBody: String? = nil,
                                 triggers: [Trigger] = []) -> Campaign {
-        let messageBodyContent = messageBody != nil ? messageBody : """
-                    {\"UIElement\" : \"\(targetViewID ?? TooltipViewIdentifierMock)\", \"position\": \"\(position.rawValue)\", \"auto-disappear\": \(autoCloseSeconds), \"redirectURL\": \"\(redirectURL)\"}
-                    """ // swiftlint:disable:previous line_length
-        return Campaign(
+        Campaign(
             data: CampaignData(
                 campaignId: id,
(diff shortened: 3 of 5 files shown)


Sources/RInAppMessaging/Router.swift
```diff
@@ -110,11 +109,15 @@ internal class Router: RouterType, ViewListenerObserver {
             return
         }
 
+        guard let presenter = getPresenter(for: campaignViewType) else {
```
imaginaris, line 112: Any reason why this is outside `displayQueue.async`?

Sources/RInAppMessaging/Router.swift
```diff
@@ -335,9 +341,6 @@ internal class Router: RouterType, ViewListenerObserver {
     }
 
     private func updateFrame(targetView: UIView, tooltipView: TooltipView, superview: UIView, position: TooltipBodyData.Position) {
-        guard targetView.superview != nil else {
```
imaginaris, line 338: Why this guard was removed?

Tests/Tests/Helpers/TestHelpers.swift
```diff
@@ -81,7 +81,9 @@ struct TestHelpers {
                                 maxImpressions: Int = 2,
                                 autoCloseSeconds: Int = 0,
                                 redirectURL: String = "",
-                                triggers: [Trigger] = []) -> Campaign {
+                                position: String = "top-center",
```
imaginaris: Can we use `TooltipBodyData.Position` instead of String?

Tests/Tests/RouterSpec.swift
```diff
@@ -108,6 +114,38 @@ class RouterSpec: QuickSpec {
                             .to(containElementSatisfying({ $0 is ModalView }))
                     }
 
+                    it("will throw assertion if the campaign presenter cannot found") {
```
imaginaris: It should be "...cannot be found". (Same for test names below)

Tests/Tests/RouterSpec.swift
```diff
@@ -139,6 +177,11 @@ class RouterSpec: QuickSpec {
                     expect(window.findIAMView()).toEventually(beNil())
                 }
 
+                it("will not remove the campaign that doesn't exist") {
```
imaginaris: I'm not sure I understand the context here. You try to call `discardDisplayedCampaign()` when no campaign is being displayed?

Tests/Tests/RouterSpec.swift
```diff
@@ -217,6 +260,45 @@ class RouterSpec: QuickSpec {
...
+                                          completion: { _ in })
+                    expect(window.findTooltipView()).toEventuallyNot(beNil())
+                    expect(router.isDisplayingTooltip(with: TooltipViewIdentifierMock)).to(beTrue())
+                }
+
+                it("will return false when tooltip is not displayed") {
+                    expect(router.isDisplayingTooltip(with: "test")).to(beFalse())
+                }
+
+                it("will return a value when calling the function from main thread") {
```
imaginaris: What value is returned? You don't call anything on a main thread here

Tests/Tests/RouterSpec.swift
```diff
@@ -217,6 +260,45 @@ class RouterSpec: QuickSpec {
...
+                    let q = DispatchQueue(label: "test")
+                    q.async {
+                        expect(router.isDisplayingTooltip(with: "test")).to(beFalse())
```
imaginaris: This will not be executed because test will not wait for this thread. You need to use `waitUntil` block and call `done()` after `expect`

Tests/Tests/RouterSpec.swift
```diff
@@ -279,6 +361,38 @@ class RouterSpec: QuickSpec {
                         expect(window.findTooltipView()).toEventuallyNot(beNil())
                     }
 
+                    it("will display a tooltip above the campaign view") {
```
imaginaris: What does it mean "above the campaign view". Tooltips should be displayed "under" a campaign view. To check that you need to check the index of tooltip view

Tests/Tests/RouterSpec.swift
```diff
@@ -279,6 +361,38 @@ class RouterSpec: QuickSpec {
...
+                                              becameVisibleHandler: { _ in },
+                                              confirmation: true,
+                                              completion: { _ in })
+                        expect(window.findTooltipView()).toEventuallyNot(beNil())
+                        window.findTooltipView()?.removeFromSuperview()
+                        window.findIAMView()?.removeFromSuperview()
+                    }
+
+                    it("will not display a tooltip campaign with no tooltip data") {
+                        let tooltip = TestHelpers.generateTooltip(id: "test", isTooltip: false)
```
imaginaris: You can use `generateCampaign` instead. `isTooltip` parameter doesn't make sense with `generateTooltip` method. How can you generate a tooltip that is not a tooltip?

Tests/Tests/RouterSpec.swift
```diff
@@ -411,6 +537,24 @@ class RouterSpec: QuickSpec {
                         scrollView.removeFromSuperview()
                     }
 
+                    it("will insert a tooltip in the same UIWindow type view as the targetView") {
+                        let uiWindow = UIWindowSubclass()
```
imaginaris, line 562: a normal `UIView` wouldn't be enough for this test?

Tests/Tests/RouterSpec.swift
```diff
@@ -450,6 +594,87 @@ class RouterSpec: QuickSpec {
                         expect(displayedTooltip.frame.origin).toEventually(equal(lastTooltipPosition.applying(translation)))
                     }
 
+                    it("will show tooltip's position in the correct position") {
```
imaginaris: This test is unnecessary. UI behaviour should be tested in UI tests but in case of tooltip positioning it's just something to be tested in QA tests. Testing position is difficult and brings a big chunk of SDK logic copied into test spec. No unit test should rely on UI Constants defined in SDK classes.
Please remove this test OR just check if `displayedTooltip` frame is different for each position - that's enough.

Sources/RInAppMessaging/Models/Responses/Campaign.swift
```diff
@@ -52,10 +52,10 @@ internal struct Campaign: Codable, Hashable {
         try container.encode(isOptedOut, forKey: .isOptedOut)
     }
 
-    init(data: CampaignData) {
+    init(data: CampaignData, asTooltip isTooltip: Bool = true) {
```
imaginaris: I don't like this new parameter. It's just aborting tooltip data parsing making the logic unnecessarily complicated just for unit tests. Please remove that flag and use either invalid tooltip data payload or just a normal campaign to make `tooltipData` nil

imaginaris (changes requested):

Sources/RInAppMessaging/Router.swift
ts-robih-robihamanto: If we move this inside `displayQueue.async {` the assertionFailure() will be handled inside displayQueue.async block and cannot be tested.
Also, I refer to how's `displayTooltip` works when finding **presenter** first before it can go through the below code.

Sources/RInAppMessaging/Router.swift
ts-robih-robihamanto: This targetView.superview will never be nil because it's already check in Line 215: 
```swift
let superview = self.findParentViewForTooltip(targetView)
```
When the superview is nil, it will return an error.

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: Yes, to cover the failure case on `guard` check.

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: I want to cover the UIWindow case on this switch case
```swift
switch superview {
        case is UIScrollView:
            return superview
        case is UIWindow:
            guard accessibilityCompatibleDisplay,
               let transitionViewClass = NSClassFromString("UITransitionView"),
               let transitionView = superview.subviews.first(where: { !$0.isKind(of: transitionViewClass) }) else {
                return superview
            }
            return transitionView
        default:
            return findParentViewForTooltip(superview)
        }
```

The UIView will not cover that case.

Sources/RInAppMessaging/Models/Responses/Campaign.swift
ts-robih-robihamanto: I agree with you, but now the only way to make Campaign instance is by passing **CampaignData to the constructor.**
Invalid tooltip data payload cannot be achieved because TooltipData inside the CampaignData is required and never be nil.

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: Same comment [here](https://github.com/rakutentech/ios-inappmessaging/pull/275#discussion_r1178695652), I think we need to find solution for this issue.

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: Because we need to cover each position in `updateFrame` function, I changed the case to check whether the tooltip is displayed in a different position.

Sources/RInAppMessaging/Models/Responses/Campaign.swift
ts-robih-robihamanto: @imaginaris 
I remove the additional params in the constructor on this commit [1ca5512](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/1ca55121edaadfeb99f5463f0caf2a453f7aba1c).

Since the `MessagePayload` is required and will never be nil. I am thinking why the TooltipData is `optional?` because it will always have data even if the MessagePayload is invalid.

The solution I am thinking of is like the below approach:

1. Remove the optional in this code.
```swift
// Campaign.swift | L.12
let tooltipData: TooltipData?
```

2. Add a new constructor for tooltip data.
```swift
    init(data: CampaignData) {
        self.data = data
        impressionsLeft = data.maxImpressions
        tooltipData = nil
    }

    init(data: CampaignData, tooltipData: MessagePayload) {
        self.data = data
        impressionsLeft = data.maxImpressions
        self.tooltipData = Campaign.parseTooltipData(messagePayload: tooltipData)
    }
```

Tests/Tests/Helpers/TestHelpers.swift
ts-robih-robihamanto: I update this here [f60c555](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/f60c555517519f44c4bd47352520a9940b80058f).

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: Fixed here [f60c555](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/f60c555517519f44c4bd47352520a9940b80058f).

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: Updated here [f60c555](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/f60c555517519f44c4bd47352520a9940b80058f).

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: Yes, you are correct. I want to handle the case when the tooltip is shown above the campaign, the campaign should bring to the front. I update here [f60c555](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/f60c555517519f44c4bd47352520a9940b80058f).

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: I meant it keeps returning the correct value even if calling it from a background thread. I update the comment here [fb6f90b](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/fb6f90b38ac0259a62a984c0e0ab078caa71ced2).

ts-dzmitry-kurski (approved):

Sources/RInAppMessaging/Router.swift
```diff
@@ -43,7 +43,7 @@ internal protocol RouterType: ErrorReportable {
 /// Handles all the displaying logic of the SDK.
 internal class Router: RouterType, ViewListenerObserver {
 
-    private enum UIConstants {
+    enum UIConstants {
```
imaginaris: please make it private again

Sources/RInAppMessaging/Router.swift
imaginaris: Ok. I think it should be fine

Sources/RInAppMessaging/Router.swift
```diff
@@ -132,46 +135,52 @@ internal class Router: RouterType, ViewListenerObserver {
...
+            return nil
+        }
+        return presenter
+    }
+
+    private func createViewConstructor(for campaign: Campaign, presenter: BaseViewPresenterType, associatedImageData: Data?) -> (() -> BaseView) {
+        var view: (() -> BaseView)!
+        let type = campaign.data.type
+
+        if type == .modal || type == .full {
```
imaginaris: `switch` would be more preferable here. Also, even if it's unrealistic, we need to handle the case of `.invalid` and `.html` types. Currently this function will crash in case of any of those types. (Also in the future we will add html type support)

Tests/Tests/RouterSpec.swift
imaginaris: Ok. I think to make it more sensible please rename this test to "will not crash when there are no displayed campaigns" AND remove `expect(window.findIAMView()).toEventually(beNil())` line. (Leaving only `router.discardDisplayedCampaign()`)

Tests/Tests/RouterSpec.swift
```diff
@@ -217,6 +260,48 @@ class RouterSpec: QuickSpec {
...
+                                          confirmation: true,
+                                          completion: { _ in })
+                    expect(window.findTooltipView()).toEventuallyNot(beNil())
+                    expect(router.isDisplayingTooltip(with: TooltipViewIdentifierMock)).to(beTrue())
+                }
+
+                it("will return false when tooltip is not displayed") {
+                    expect(router.isDisplayingTooltip(with: "test")).to(beFalse())
+                }
+
```
imaginaris, line 292: please add one more case when tooltip is displayed but the function is called with another identifier.

Tests/Tests/RouterSpec.swift
```diff
@@ -217,6 +260,48 @@ class RouterSpec: QuickSpec {
...
+                it("will keep return Bool value when calling the from background thread") {
```
imaginaris, line 304: Is this test checking if there's no thread lock?
In that case you can just say "will return a value ..." in the beginning

Tests/Tests/RouterSpec.swift
```diff
@@ -279,6 +364,34 @@ class RouterSpec: QuickSpec {
...
+                        expect(window.findTooltipView()).toEventuallyNot(beNil())
+                        expect(window.findIAMView()).toEventuallyNot(beNil())
+                        let displayedTooltip = window.findTooltipView()!
+                        let displayedCampaign = window.findIAMView()!
+
+                        expect(displayedTooltip.superview).to(beIdenticalTo(displayedCampaign.superview))
+                        let tooltipIndex = displayedTooltip.superview?.subviews.firstIndex(of: displayedTooltip)
+                        let campaignIndex = displayedTooltip.superview?.subviews.firstIndex(of: displayedCampaign)
+                        expect(campaignIndex).to(beGreaterThan(tooltipIndex))
+                        window.findTooltipView()?.removeFromSuperview()
```
imaginaris, line 401: Is this cleanup necessary?
Tooltip removal should be handled by line 320 and campaign view by line 50.

Tests/Tests/RouterSpec.swift
```diff
@@ -279,6 +364,34 @@ class RouterSpec: QuickSpec {
...
+                                              completion: { _ in })
+                        expect(window.findTooltipView()).toEventuallyNot(beNil())
+                        expect(window.findIAMView()).toEventuallyNot(beNil())
+                        let displayedTooltip = window.findTooltipView()!
+                        let displayedCampaign = window.findIAMView()!
+
+                        expect(displayedTooltip.superview).to(beIdenticalTo(displayedCampaign.superview))
+                        let tooltipIndex = displayedTooltip.superview?.subviews.firstIndex(of: displayedTooltip)
+                        let campaignIndex = displayedTooltip.superview?.subviews.firstIndex(of: displayedCampaign)
+                        expect(campaignIndex).to(beGreaterThan(tooltipIndex))
```
imaginaris, line 400: 👍

Tests/Tests/RouterSpec.swift
```diff
@@ -450,6 +593,34 @@ class RouterSpec: QuickSpec {
...
+                            let tooltip = TestHelpers.generateTooltip(id: "test", position: position)
+                            router.displayTooltip(tooltip,
+                                                  targetView: targetView,
+                                                  identifier: TooltipViewIdentifierMock,
+                                                  imageBlob: imageBlob,
+                                                  becameVisibleHandler: { _ in },
+                                                  confirmation: true,
+                                                  completion: { _ in })
+                            expect(window.findTooltipView()).toEventuallyNot(beNil())
+                            let displayedTooltip = window.findTooltipView()!
```
imaginaris, line 640: It's a coincidence that `findTooltipView()` returns the most recently displayed tooltip. Before calling this function you should remove the displayed tooltip from the previous loop to ensure accurate and stable test result.

Sources/RInAppMessaging/Models/Responses/Campaign.swift
imaginaris: Hmm how can you remove optional from here `let tooltipData: TooltipData?` and then have this line in the constructor: `tooltipData = nil`?
`Campaign.parseTooltipData(messagePayload: tooltipData)` can return nil and it will when message body doesn't contain a valid json file that can be mapped to `TooltipData` model.
The only way to distinguish tooltip payload from normal campaign payload is by checking the content of `messageBody`. Anyway, the main point is that it's a temporary measure. The tooltip feature is in an alpha phase. The data model parsing will change when IAM Dashboard start supporting tooltip campaign type.

imaginaris: > Invalid tooltip data payload cannot be achieved because TooltipData inside the CampaignData is required and never be nil.

What do you mean by 'TooltipData inside the CampaignData'?
You can have invalid tooltip data payload by putting an empty string in `messageBody` or a json with a missing key that is required.
(ref: https://github.com/rakutentech/ios-inappmessaging/blob/master/Tests/Tests/SerializationSpec.swift#L126)

Sources/RInAppMessaging/Router.swift
ts-robih-robihamanto: Updated here [57d4646](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/57d464617318d1c37bbc0e36be3d5a49f81e4b83).

Sources/RInAppMessaging/Router.swift
ts-robih-robihamanto: Updated here [57d4646](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/57d464617318d1c37bbc0e36be3d5a49f81e4b83)

Sources/RInAppMessaging/Models/Responses/Campaign.swift
ts-robih-robihamanto: Yes, you are correct and thank you for sending the ref. I finally can achieve the invalid tooltipData with `""` messageBody. I update the code here [1277742](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/127774286b2c453a2a3c279fc8a1d61dcb388b3c).

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: This case is checking whether the function will work properly when the function is called in a background thread.
It will cover this line.
image

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: Yes, it's because of the `rootView` as the targetView we need to remove it from the `rootView`.

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: Update here [3cee1ec](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/3cee1ec04fdbedfab437064f59c968abeb121232).

Tests/Tests/RouterSpec.swift
ts-robih-robihamanto: Addded here [3cee1ec](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/3cee1ec04fdbedfab437064f59c968abeb121232).

Tests/Tests/Helpers/TestHelpers.swift
```diff
@@ -81,10 +81,13 @@ struct TestHelpers {
...
                                 autoCloseSeconds: Int = 0,
                                 redirectURL: String = "",
-                                position: String = "top-center",
-                                triggers: [Trigger] = [],
-                                isTooltip: Bool = true) -> Campaign {
-        Campaign(
+                                position: TooltipBodyData.Position = .topCenter,
+                                messageBody: String? = nil,
+                                triggers: [Trigger] = []) -> Campaign {
+        let messageBodyContent = messageBody != nil ? messageBody : """
```
imaginaris: Can you use `??` operator instead? I mean like this:
```
let messageBodyContent = messageBody ?? """
{\"UIElement\" : \"\(targetViewID ?? TooltipViewIdentifierMock)\", \"position\": \"\(position.rawValue)\", \"auto-disappear\": \(autoCloseSeconds), \"redirectURL\": \"\(redirectURL)\"}
"""
```

Tests/Tests/RouterSpec.swift
```diff
@@ -217,6 +259,52 @@ class RouterSpec: QuickSpec {
...
+                    expect(window.findTooltipView()).toEventuallyNot(beNil())
+                    expect(router.isDisplayingTooltip(with: TooltipViewIdentifierMock)).to(beTrue())
+                }
+
+                it("will return false when tooltip is not displayed") {
+                    expect(router.isDisplayingTooltip(with: "test")).to(beFalse())
+                }
+
+                it("will return false when tooltip identifier is incorrect") {
+                    expect(router.isDisplayingTooltip(with: "test")).to(beFalse())
```
imaginaris: But you don't display any tooltip here.
You need to have a code like this before the final expectation:
```
                    router.displayTooltip(tooltip, targetView: tooltipTargetView,
                                          identifier: TooltipViewIdentifierMock,
                                          imageBlob: imageBlob,
                                          becameVisibleHandler: { _ in },
                                          confirmation: true,
                                          completion: { _ in })
                    expect(window.findTooltipView()).toEventuallyNot(beNil())
```

ts-robih-robihamanto: Updated here [72da23d](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/72da23dfdd6189fdffe59dbeb76b66e340dc6f9c).

Tests/Tests/RouterSpec.swift
```diff
@@ -291,7 +291,13 @@ class RouterSpec: QuickSpec {
...
 
                 it("will return false when tooltip identifier is incorrect") {
-                    expect(router.isDisplayingTooltip(with: "test")).to(beFalse())
+                    router.displayTooltip(tooltip, targetView: tooltipTargetView,
+                                          identifier: TooltipViewIdentifierMock,
+                                          imageBlob: imageBlob,
+                                          becameVisibleHandler: { _ in },
+                                          confirmation: true,
+                                          completion: { _ in })
+                    expect(router.isDisplayingTooltip(with: "incorrect_id")).to(beFalse())
```
imaginaris, line 301: You need to wait for tooltip to appear first. i.e. put this `expect(window.findTooltipView()).toEventuallyNot(beNil())` before `expect()`

ts-robih-robihamanto: I missed this part, added [1a77971](https://github.com/rakutentech/ios-inappmessaging/pull/275/commits/1a77971849da0b88654dddd237c75cf2d33c9216).

imaginaris (approved):

PR #279, opened 2023-05-18
feat: add SwiftUI support for Tooltips (SDKCF-5025)

imaginaris:
# Description
Unit tests in a separate PR

## Links
SDKCF-5025

# Checklist
- [X] All project file changes are replicated in `SampleSPM/SampleSPM.xcodeproj` project

### DIFF
--- a/Sources/RInAppMessaging/Extensions/View+IAM.swift
+++ b/Sources/RInAppMessaging/Extensions/View+IAM.swift
@@ -1,0 +1,30 @@
+import Foundation
+import SwiftUI
+
+@available(iOS 13.0, *)
+extension View {
+    /// Hide or show the view based on a boolean value.
+    ///
+    /// Example for visibility:
+    ///
+    ///     Text("Label")
+    ///         .isHidden(true)
+    ///
+    /// Example for complete removal:
+    ///
+    ///     Text("Label")
+    ///         .isHidden(true, remove: true)
+    ///
+    /// - Parameters:
+    ///   - hidden: Set to `false` to show the view. Set to `true` to hide the view.
+    ///   - remove: Boolean value indicating whether or not to remove the view.
+    @ViewBuilder func isHidden(_ hidden: Bool, remove: Bool = false) -> some View {
+        if hidden {
+            if !remove {
+                self.hidden()
+            }
+        } else {
+            self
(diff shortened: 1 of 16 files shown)


Sources/RInAppMessaging/Router.swift
```diff
@@ -43,7 +50,7 @@ internal protocol RouterType: ErrorReportable {
 /// Handles all the displaying logic of the SDK.
 internal class Router: RouterType, ViewListenerObserver {
 
-    private enum UIConstants {
```
ra-malik-alayli, line 46: @imaginaris What is the reason of removing `private`?

Sources/RInAppMessaging/Extensions/View+IAM.swift
```diff
@@ -0,0 +1,30 @@
...
+    ///   - hidden: Set to `false` to show the view. Set to `true` to hide the view.
+    ///   - remove: Boolean value indicating whether or not to remove the view.
+    @ViewBuilder func isHidden(_ hidden: Bool, remove: Bool = false) -> some View {
+        if hidden {
+            if !remove {
+                self.hidden()
+            }
+        } else {
+            self
+        }
```
ra-malik-alayli, line 28: Maybe using guard instead of if/else could be more readable here.

Sources/RInAppMessaging/Router.swift
imaginaris: Those constants are accessed by `TooltipViewModifier`. Actually I should move this enum

Sources/RInAppMessaging/Extensions/View+IAM.swift
imaginaris: True (blind copy&paste)

imaginaris: Ah I forgot. this is a ViewBuilder function. You can't use guard so it has to look like this. SwiftUI views can become unreadable unfortunately when conditions get involved.
(`Closure containing control flow statement cannot be used with result builder 'ViewBuilder'`)

ra-malik-alayli (approved):

ts-dzmitry-kurski (approved):

PR #283, opened 2023-05-28
fix: fix Tooltip event handling (SDKCF-5025)

imaginaris:
# Description
This PR fixes issue with tooltip campaigns appearing too many times (UIKit) or not appearing again when view was reloaded (SwiftUI)

## Links
SDKCF-5025

# Checklist
- [X] All project file changes are replicated in `SampleSPM/SampleSPM.xcodeproj` project

### DIFF
--- a/Sources/RInAppMessaging/Events/ViewAppearedEvent.swift
+++ b/Sources/RInAppMessaging/Events/ViewAppearedEvent.swift
@@ -3,7 +3,7 @@ import Foundation
 /// A special event for internal handling of Tooltip display cycle
 internal class ViewAppearedEvent: Event {
 
-    let viewIdentifier: String
+    internal let viewIdentifier: String
 
     init(viewIdentifier: String) {
         self.viewIdentifier = viewIdentifier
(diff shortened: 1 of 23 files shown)


ts-dzmitry-kurski (approved):

Sources/RInAppMessaging/Events/ViewAppearedEvent.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 /// A special event for internal handling of Tooltip display cycle
 internal class ViewAppearedEvent: Event {
 
-    let viewIdentifier: String
+    internal let viewIdentifier: String
```
Esakkiraja-Pothikannan, line 6: Is there any specific reason to explicitly mark it as `internal` here? 
(I guess that this is just to clue others dev to be careful about changing it to public/open)

imaginaris: The only reason is to make it consistent with the rest of the codebase. In general, putting `internal` is rather redundant since it's an implicit declaration. We don't use that in other projects.
The only advantage I can think of is to make user/developer aware that this is definitely not a part of public API.

PR #297, opened 2023-08-10
feat: ignore `configure()` calls when RMC module is integrated (SDKCF-6710)

imaginaris:
# Description
When RMC module is integrated, allow only `configure()` calls that come from RMC source code.

Note:
I couldn't find a way to write unit tests for this feature without complicating the code.

## Links
SDKCF-6710

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Constants.swift
+++ b/Sources/RInAppMessaging/Constants.swift
@@ -2,8 +2,6 @@ import typealias Foundation.TimeInterval
 
 internal enum Constants {
 
-    static let rmcSubscriptionIDSuffix = "-rmc"
-
     enum CampaignMessage {
         static let imageRequestTimeoutSeconds: TimeInterval = 20
         static let imageResourceTimeoutSeconds: TimeInterval = 300
@@ -62,6 +60,10 @@ internal enum Constants {
             static let pushPermission = "push_permission"
             static let deviceID = "device_id"
         }
+    }
+
+    enum RMC {
+        static let subscriptionIDSuffix = "-rmc"
     }
 
     enum Retry {
--- a/Sources/RInAppMessaging/Constants.swift
+++ b/Sources/RInAppMessaging/Constants.swift
@@ -2,6 +2,8 @@ import typealias Foundation.TimeInterval
 
 internal enum Constants {
 
+    static let rmcSubscriptionIDSuffix = "-rmc"
+
     enum CampaignMessage {
         static let imageRequestTimeoutSeconds: TimeInterval = 20
         static let imageResourceTimeoutSeconds: TimeInterval = 300
--- a/Sources/RInAppMessaging/RInAppMessaging.swift
+++ b/Sources/RInAppMessaging/RInAppMessaging.swift
@@ -18,6 +18,10 @@ import RSDKUtils
     internal static var swiftUIEventHandler: SwiftUIViewEventHandlerType? {
         dependencyManager?.resolve(type: SwiftUIViewEventHandlerType.self)
     }
+    /// Returns `true` when RMC module is integrated in the host app
+    internal static var isRMCEnvironment: Bool {
+        NSClassFromString("RMCInAppMessaging.RMCInAppMessaging") != nil
+    }
 
     private override init() { super.init() }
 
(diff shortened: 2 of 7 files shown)


malayli (approved):

Sources/RInAppMessaging/RInAppMessaging.swift
```diff
@@ -60,11 +60,14 @@ import RSDKUtils
...
     ///     - configurationURL: a configuration URL. (This setting will override the `InAppMessagingConfigurationURL` value in Info.plist)
-    ///     - enableTooltipFeature: set to `true` to enable Tooltip campaigns. This feature is currently in beta phase. Default value: `false`
+    ///     - enableTooltipFeature: set to `true` to enable Tooltip campaigns. This feature is currently in beta phase. Default value: `false`.
+    ///     - caller: for internal use only.
     @objc public static func configure(subscriptionID: String? = nil,
                                        configurationURL: String? = nil,
-                                       enableTooltipFeature: Bool = false) {
-        guard initializedModule == nil else {
+                                       enableTooltipFeature: Bool = false,
+                                       caller: String = #file) {
```
maureenorea-clores: It is not a breaking change, but is it okay to update public API?

Esakkiraja-Pothikannan: @imaginaris Can we follow the same approach followed by Android?

imaginaris: @maureenorea-clores I couldn't find any other way to identify if the `configure()` is called by RMC or not.
I think it's ok as long as we don't break the API.

Sources/RInAppMessaging/RInAppMessaging.swift
```diff
@@ -211,6 +214,17 @@ import RSDKUtils
         return configURL
     }
 
+    /// Checks the existence of RMC module and verifies the `configure()` caller.
+    /// - Parameter caller: a path to the source file that called `configure()`
+    /// - Returns: `false` if RMC module is integrated and the method wasn't called from the RMC module.
+    private static func verifyRMCEnvironment(caller: String) -> Bool {
```
Esakkiraja-Pothikannan: Please make it `internal static func` without any function param so that we can use this func for SDKCF-6709

Sources/RInAppMessaging/RInAppMessaging.swift
imaginaris: @maureenorea-clores Or actually we have just decided to use the same approach for RMC identification used in Android 😬 . So the public API will not be modified

maureenorea-clores: @imaginaris I see, okay thanks

Sources/RInAppMessaging/RInAppMessaging.swift
```diff
@@ -18,6 +18,10 @@ import RSDKUtils
     internal static var swiftUIEventHandler: SwiftUIViewEventHandlerType? {
         dependencyManager?.resolve(type: SwiftUIViewEventHandlerType.self)
     }
+    /// Returns `true` when RMC module is integrated in the host app
+    internal static var isRMCEnvironment: Bool {
+        NSClassFromString("RMCInAppMessaging.RMCInAppMessaging") != nil
```
Esakkiraja-Pothikannan: `RMCInAppMessaging` is not a class.  Please update RMCSDK.

Sources/RInAppMessaging/Constants.swift
```diff
@@ -2,6 +2,8 @@ import typealias Foundation.TimeInterval
 
 internal enum Constants {
 
+    static let rmcSubscriptionIDSuffix = "-rmc"
```
Esakkiraja-Pothikannan: Add suffix key in RMCSDK.

Sources/RInAppMessaging/RInAppMessaging.swift
imaginaris: done

Tests/Tests/PublicAPISpec.swift
```diff
@@ -659,6 +659,19 @@ class PublicAPISpec: QuickSpec {
...
             }
+
+            context("when calling sanitizeSubscriptionID()") {
+
+                it("will remove '-rmc' suffix") {
+                    let subscriptionID = "myKey-rmc"
+                    expect(RInAppMessaging.sanitizeSubscriptionID(subscriptionID)).toEventually(equal("myKey"))
+                }
+
+                it("will not remove '-rmc' substring if it's not at the and of the string") {
```
maureenorea-clores: typo: and

maureenorea-clores (approved):

Sources/RInAppMessaging/RInAppMessaging.swift
```diff
@@ -211,6 +216,30 @@ import RSDKUtils
...
 
+    /// Checks the existence of RMC module and verifies the `configure()` caller.
+    /// - Parameter subscriptionID: a subscriptionID value from `configure()` call to check for '-rmc' suffix
+    /// - Returns: `false` if RMC module is integrated and the method wasn't called from the RMC module.
+    internal static func verifyRMCEnvironment(subscriptionID: String?) -> Bool {
+        guard isRMCEnvironment else {
+            return true
+        }
+
+        return subscriptionID?.hasSuffix(Constants.RMC.subscriptionIDSuffix) == true
```
Esakkiraja-Pothikannan, line 227: Could be better if use `nil coalescing` like `?? false`

Esakkiraja-Pothikannan (approved):

Sources/RInAppMessaging/RInAppMessaging.swift
imaginaris: @Esakkiraja-Pothikannan But how? `(subscriptionID ?? "").hasSuffix(Constants.RMC.subscriptionIDSuffix) == true`?
Only `subscriptionID` is optional here

Esakkiraja-Pothikannan: I meant this way `subscriptionID?.hasSuffix(Constants.RMC.subscriptionIDSuffix) ?? false`

imaginaris: Ah ok. Interesting. Not sure which looks better. I think mine is more straightforward to be honest.

Sources/RInAppMessaging/Constants.swift
imaginaris: Created a PR in RMC SDK

PR #317, opened 2024-04-30
fix: Prevent sending impressions for visible but expired campaigns (SDKCF-6910)

SoumenRautray:
# Description
RCA:
When campaigns are not dismissed even after they are expired, on closing the campaign impressions requests are being sent. So Impressions are seen for expired campaigns which is not the expected behaviour 

Fix:
Validated the campaign if its outdated while sending the impression. Impressions for outdated campaigns will not be sent now

## Links
[SDKCF-6910]

# Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Changelog
 
 ### Unreleased
+- Improvements:
+	- Prevent sending impressions for expired campaigns [SDKCF-6910]
 
 ### 8.2.0 (2024-04-08) 
 - Improvements:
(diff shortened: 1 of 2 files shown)


maureenorea-clores (commented):
Yes the impression will not be sent, but user still able to see the expired campaign. Is it accepted behavior?

SoumenRautray:
> Yes the impression will not be sent, but user still able to see the expired campaign. Is it accepted behavior?

Yes, its expected behaviour as per the the discussions with PDMs and @Esakkiraja-Pothikannan

maureenorea-clores (approved):

ramesh-kuntamala (approved):
Looks good

PR #333, opened 2024-09-24
chore : Implement dark background through CustomJson (RMCCX-7328)

SoumenRautray:
# Description
Implement dark background through CustomJson

## Links
[RMCCX-7328]

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Models/Responses/CampaignDataModels.swift
+++ b/Sources/RInAppMessaging/Models/Responses/CampaignDataModels.swift
@@ -95,7 +95,7 @@ internal struct CustomJson: Codable {
     }
 }
 
-internal struct PrimerButton: Codable {
+internal struct PrimerButton: Codable, Equatable {
     let button: Int?
     
     enum CodingKeys: String, CodingKey {
@@ -112,7 +112,7 @@ internal struct PrimerButton: Codable {
     }
 }
 
-internal struct ClickableImage: Codable {
+internal struct ClickableImage: Codable, Equatable{
     let url: String?
     
     enum CodingKeys: String, CodingKey {
@@ -129,7 +129,7 @@ internal struct ClickableImage: Codable {
     }
 }
 
-internal struct BackgroundColor: Codable {
+internal struct BackgroundColor: Codable, Equatable {
     let opacity: Double?
     
     enum CodingKeys: String, CodingKey {
--- a/Sources/RInAppMessaging/Models/Responses/CampaignDataModels.swift
+++ b/Sources/RInAppMessaging/Models/Responses/CampaignDataModels.swift
@@ -73,21 +73,25 @@ internal struct ButtonBehavior: Codable {
 internal struct CustomJson: Codable {
     let pushPrimer: PrimerButton?
     let clickableImage: ClickableImage?
+    let background: BackgroundColor?
     
     enum CodingKeys: String, CodingKey {
         case pushPrimer
         case clickableImage
+        case background
     }
     
-    init(pushPrimer: PrimerButton? = nil, clickableImage: ClickableImage? = nil) {
+    init(pushPrimer: PrimerButton? = nil, clickableImage: ClickableImage? = nil, background: BackgroundColor? = nil) {
         self.pushPrimer = pushPrimer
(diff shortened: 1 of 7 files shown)


Sources/RInAppMessaging/Models/Responses/CampaignDataModels.swift
```diff
@@ -132,3 +136,24 @@ internal struct ClickableImage: Codable {
...
+        case opacity
+    }
+    
+    init(opacity: Double? = nil) {
+        self.opacity = opacity
+    }
+    
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        if let opacity = try? container.decode(Double.self, forKey: .opacity) {
```
Esakkiraja-Pothikannan: Could be written like this `self.opacity = try? container.decode(Double.self, forKey: .opacity)` instead of `if let ... else`

Sources/RInAppMessaging/Models/Responses/CampaignDataModels.swift
```diff
@@ -73,21 +73,25 @@ internal struct ButtonBehavior: Codable {
...
         self.pushPrimer = pushPrimer
         self.clickableImage = clickableImage
+        self.background = background
     }
     
     init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
         pushPrimer = try container.decodeIfPresent(PrimerButton.self, forKey: .pushPrimer)
         clickableImage = try container.decodeIfPresent(ClickableImage.self, forKey: .clickableImage)
+        background = try container.decodeIfPresent(BackgroundColor.self, forKey: .background)
```
Esakkiraja-Pothikannan: What if the data type of background is String (not dictionary)?

Tests/Tests/ViewSpec.swift
```diff
@@ -174,6 +174,57 @@ class ViewSpec: QuickSpec {
...
+                expect(view.backgroundViewColor).to(equal(.black.withAlphaComponent(0)))
+
+                view.setup(viewModel: TestHelpers.generateFullViewModel(customJson: CustomJson(background: BackgroundColor(opacity: 1))))
+                expect(view.backgroundViewColor).to(equal(.black.withAlphaComponent(1)))
+            }
+
+            it("will have default background color if CustomJson has invalid value") {
+                view.setup(viewModel: TestHelpers.generateFullViewModel(customJson: CustomJson(background: BackgroundColor(opacity: 1.2))))
+                expect(view.backgroundViewColor).toNot(equal(.black.withAlphaComponent(1.2)))
+                expect(view.backgroundViewColor).to(equal(.black.withAlphaComponent(0.14)))
```
Esakkiraja-Pothikannan, line 209: why this equal to `.black.withAlphaComponent(0.14)` for invalid opacity value?

Esakkiraja-Pothikannan (commented):
We should have test to cover the parsing logic of customJson with `background` by passing valid & invalid values.

Tests/Tests/ViewSpec.swift
SoumenRautray: This the the default color and opacity for all the campaigns , so if the custom json does not have valid value then it will have the default color and opacity

Esakkiraja-Pothikannan: i.e it will show black background with opacity of 0.14 for all campaign by default not a clear colour, is it?

Sources/RInAppMessaging/Models/Responses/CampaignDataModels.swift
SoumenRautray: try? can be used to return nil for this scenario

Tests/Tests/ViewSpec.swift
SoumenRautray: yes, uiConstants.backgroundColor ?? viewModel.backgroundColor  this the default color 
var backgroundViewColor: UIColor? = .clear was used for testing purpose

Esakkiraja-Pothikannan (approved):

ramesh-kuntamala (approved):

PR #341, opened 2024-12-20
chore: Implement autoscroll and add user click behavior for Carousel (RMCCX-7618, RMCCX-7620)

SoumenRautray:
# Description
Implemented autoscroll and add user click behavior for Carousel 

## Links
[RMCCX-7618]
[RMCCX-7620]

# Checklist

### DIFF
--- a/Sources/RInAppMessaging/Views/CarouselView.swift
+++ b/Sources/RInAppMessaging/Views/CarouselView.swift
@@ -6,6 +6,11 @@ import UIKit
     @IBOutlet weak var carouselHeightConstraint: NSLayoutConstraint!
 
     var carouselData: [CarouselData] = []
+    private var timer: Timer?
+    private var userIsScrolling = false
+    private var currentIndex = 0
+    private var hasReachedLastImage = false
+    var presenter: FullViewPresenterType?
 
     required init?(coder: NSCoder) {
         super.init(coder: coder)
@@ -14,12 +19,15 @@ import UIKit
 
     deinit {
         NotificationCenter.default.removeObserver(self, name: UIDevice.orientationDidChangeNotification, object: nil)
+        stopAutoScroll()
     }
 
-    func configure(carouselData: [CarouselData]) {
+    func configure(carouselData: [CarouselData], presenter: FullViewPresenterType) {
         self.carouselData = carouselData
+        self.presenter = presenter
         setupCollectionView()
         setupPageControl()
+        startAutoScroll()
         collectionView.reloadData()
     }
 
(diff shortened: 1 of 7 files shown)


Sources/RInAppMessaging/Views/CarouselView.swift
```diff
@@ -122,6 +138,57 @@ extension CarouselView {
...
+
+        let nextIndex = currentIndex + 1
+        let indexPath = IndexPath(item: nextIndex, section: 0)
+
+        collectionView.scrollToItem(at: indexPath, at: .centeredHorizontally, animated: true)
+        currentIndex = nextIndex
+    }
+
+    func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
+        userIsScrolling = true
```
ramesh-kuntamala: I don't think userIsScrolling property is required because when you invalidate and make timer as nil, auto scroll will be stopped

SoumenRautray: yes, that is the purpose. When carousel loads it starts auto scrolling , if the user interacts and swipes back or forward it stops auto scrolling. It will start scrolling again after 5 seconds ,only if has not completely scrolled to the end once .

ramesh-kuntamala: `userIsScrolling`  Without this property also you can achieve the same result

ramesh-kuntamala (approved):

Sources/RInAppMessaging/Views/FullView.swift
```diff
@@ -304,7 +307,7 @@ internal class FullView: UIView, FullViewType, RichContentBrowsable {
         carouselView.setPageControlVisibility(isHdden: layout != .carousel)
         buttonsContainer.isHidden = !viewModel.showButtons
         optOutView.isHidden = !viewModel.showOptOut
-        optOutButtonTopSpacer.isHidden = layout != .carousel && (!buttonsContainer.isHidden || !optOutView.isHidden)
+        optOutButtonTopSpacer.isHidden = layout == .carousel && (buttonsContainer.isHidden || optOutView.isHidden)
```
Esakkiraja-Pothikannan, line 310: what is the purpose of this property? what is the condition check for this?

SoumenRautray: this is an extra spacing which i am hiding when there is carousel view , since the pageviewcontroller already adds some spacing .

Esakkiraja-Pothikannan (approved):

